PR #414, opened 2015-11-17
Create an internal ADLogger header.

RPangrle:
This was needed to be able to remove some warnings in tokenbroker without adding more public APIs.

### DIFF
--- a/ADALiOS/ADALiOS/ADBrokerKeyHelper.m
+++ b/ADALiOS/ADALiOS/ADBrokerKeyHelper.m
@@ -23,7 +23,7 @@
 #import "ADKeyChainHelper.h"
 #import <CommonCrypto/CommonCryptor.h>
 #import <Security/Security.h>
-#import "ADLogger.h"
+#import "ADLogger+Internal.h"
 
 @implementation ADBrokerKeyHelper
 
--- a/ADALiOS/ADALiOS/ADLogger+Internal.h
+++ b/ADALiOS/ADALiOS/ADLogger+Internal.h
@@ -1,0 +1,93 @@
+// Copyright © Microsoft Open Technologies, Inc.
+//
+// All Rights Reserved
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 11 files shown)


ADALiOS/ADALiOS/ADBrokerKeyHelper.m
```diff
@@ -23,7 +23,7 @@
 #import "ADKeyChainHelper.h"
 #import <CommonCrypto/CommonCryptor.h>
 #import <Security/Security.h>
-#import "ADLogger.h"
+#import "ADLogger+Internal.h"
```
weijjia: Just out of curiosity, is it like a naming convention in objc to have the "+" in the name?

ADALiOS/ADALiOS/ADLogger+Internal.h
```diff
@@ -0,0 +1,93 @@
...
+
+/*! Logs obtaining of a token. The method does not log the actual token, only its hash.
+ @param token: the token to log.
+ @param tokenType: "access token", "refresh token", "multi-resource refresh token"
+ @param expiresOn: the time when an access token will stop to be valid. Nil for refresh token types.
+ @param correlationId: In case the token was just obtained from the server, the correlation id of the call.
+ This parameter can be nil.
+ */
++ (void)logToken:(NSString*)token
+       tokenType:(NSString*)tokenType
```
weijjia: Does objc have enum? If so, it would be better to pass a enum here,

ADALiOS/ADALiOS/ADBrokerKeyHelper.m
RPangrle: Yup, that's the naming convention for categories. (Which are a way in Objective-C of extending a class, or splitting up a class's implementation across multiple files.)

ADALiOS/ADALiOS/ADLogger+Internal.h
RPangrle: We're just printing strings here, not performing any actual logic on them. This also is previous implementation. Creating an enum would be overkill.

weijjia: I see and agree then.

weijjia:
:shipit:

PR #422, opened 2015-11-24
broker-aware adal should not call broker if authority validation is turned off 

### LINKED ISSUE #397: broker-aware adal should not call broker if authority validation is turned off
When talking directly to an ADFS server the dev has to turn off authority validation. In those cases we shouldn't call broker either, as broker does not work with naked ADFS.

jasoncoolmax:
A small change to fix #397. Added the following condition before calling the broker: the authority validation must be on. See #397 for more detail.


RPangrle:
:shipit:

Issue #426, opened 2015-12-03
Unit Tests Should Not Require Network


RPangrle:
This is causing occasional Travis failures.

PR #428, opened 2015-12-07
Unit tests changed to be networkless

jasoncoolmax:
Unit tests changed to be independent of network. Web request responses are mocked out by overriding the behavior of NSURLConnection. (for bug #426)


RPangrle:
You're missing a project file change here for the added files.

jasoncoolmax:
You are right. Project file has been pushed. :)

RPangrle:
:shipit:

RPangrle:
#426

Issue #321, opened 2015-05-20
Infinite retry on server 500 error


eddiekim:
Calling into ADAL to "acquireTokenSilentWithResource" resulted in an infinite retry loop in the ADAL library never calling back to "completionBlock".

Noticed on 1.2.2 and 1.2.4

[ADLogger callback snippet below]

ADALiOS [2015-05-20 18:41:16 - 85073B08-17A0-4F8B-BB42-E93978B8E802] Error raised: 15 (Domain: ADAuthenticationErrorDomain ProtocolCode:500 Details:Server HTTP status code: 500. Full response {"error":"server_error","error_description":"AADSTS90036: Non-retryable error has occurred.\r\nTrace ID: 040579a9-79ed-42e5-9076-e114d40a2d69\r\nCorrelation ID: 85073b08-17a0-4f8b-bb42-e93978b8e802\r\nTimestamp: 2015-05-20 18:41:16Z","error_codes":[90036],"timestamp":"2015-05-20 18:41:16Z","trace_id":"040579a9-79ed-42e5-9076-e114d40a2d69","correlation_id":"85073b08-17a0-4f8b-bb42-e93978b8e802","submit_url":null,"context":null}) (errorCode=15)
ADALiOS [2015-05-20 18:41:16 - 7DD2FCC0-3A5B-40CD-B748-FAF07F7902BB] Error raised: 15 (Domain: ADAuthenticationErrorDomain ProtocolCode:500 Details:Server HTTP status code: 500. Full response {"error":"server_error","error_description":"AADSTS90036: Non-retryable error has occurred.\r\nTrace ID: 8bd52f45-eb56-4c5c-a20c-89f82aeed24b\r\nCorrelation ID: 7dd2fcc0-3a5b-40cd-b748-faf07f7902bb\r\nTimestamp: 2015-05-20 18:41:16Z","error_codes":[90036],"timestamp":"2015-05-20 18:41:16Z","trace_id":"8bd52f45-eb56-4c5c-a20c-89f82aeed24b","correlation_id":"7dd2fcc0-3a5b-40cd-b748-faf07f7902bb","submit_url":null,"context":null}) (errorCode=15)
[...]

PR #429, opened 2015-12-09
Fix the "infinite retry on server 500 error" bug

jasoncoolmax:
1.(Referring to bug #321) 
The bug is because, when ADAL is trying the multi-resource refresh token, the multi-resource refresh token is somehow modified, which subsequently makes ADAL think it hasn't tried it yet. Therefore infinite-recursive-call happens. 

2.(Referring to bug #321 and #3)
If response codes are 500 or 502, retry will be done once after 0.5 sec at the web request level. In other words, ADAuthenticationRequest is not aware of it.


RPangrle:
Maybe a half sec? Whatever number we pick will be pretty arbitrary. It seems to me that if the server is failing because it's getting continually slammed that's not going to let up for a while, so we shouldn't wait too long.

ADALiOS/ADALiOS/ADAuthenticationRequest+AcquireAssertion.h
```diff
@@ -28,6 +28,7 @@
                useAccessToken:(BOOL)useAccessToken
                 samlAssertion:(NSString*)samlAssertion
                 assertionType:(ADAssertionType)assertionType
+           retryIfServerError:(BOOL)retryIfServerError
```
RPangrle: Make this a property on the request object itself. I'm trying to minimize the amount of method parameters if at all possible.

ADALiOS/ADALiOS/ADAuthenticationRequest+AcquireAssertion.m
```diff
@@ -191,6 +202,22 @@ - (void)attemptToUseCacheItem:(ADTokenCacheStoreItem*)item
      }];//End of the refreshing token completion block, executed asynchronously.
 }
 
+- (void)retryMultiResourceRefreshToken:(ADTokenCacheStoreItem*)item
```
RPangrle: Remove this method, just have it call itself recursively.

ADALiOS/ADALiOS/ADAuthenticationRequest+AcquireToken.m
```diff
@@ -335,7 +354,7 @@ - (void) validatedAcquireTokenByRefreshToken:(NSString*)refreshToken
           additionalHeaders:nil
                  completion:^(NSDictionary *response)
      {
-         ADTokenCacheStoreItem* resultItem = (cacheItem) ? cacheItem : [ADTokenCacheStoreItem new];
+         ADTokenCacheStoreItem* resultItem = (cacheItem) ? [ADTokenCacheStoreItem initWithStoreItem:cacheItem] : [ADTokenCacheStoreItem new];
```
RPangrle: I'm still not sure why this line is needed. Creating a copy of the cache store item shouldn't have any difference on the logic of the code.

jasoncoolmax: The call stack is like:
(1)attemptToUseCacheItem()  -> (2)acquireTokenByRefreshToken() -> (3)validatedAcquireTokenByRefreshToken()

Variable *cacheItem* above is a parameter passed from (1) all the way to (3), however, such a parameter is changed in (3) because the above deleted line didn't make a copy of the parameter. In particular, when (3) set the resource property of *resultItem*, it changes the resource property of *cacheItem* as well.

Such a behaviour causes the following problem:
(1) is responsible for trying both RT and MRRT (if necessary), which is implemented in a recursive way. Function (1) assumes variable *cacheItem* should not be changed by anybody. It tells whether the cacheItem is a RT/MRRT by looking at the resource property being nil or not. If YES, it will stop the recursive call because MRRT should be the last token it try before asking user to log in again.

The above deleted line leads function (1) to misjudge whether cacheItem is RT or MRRT, therefore harm the recursive call terminating condition. In particular, it causes infinite retry for 500 error.

(PS: infinite recursive call happens in 500 error but not some other error code is because, 500 error will not cause any MRRT to be deleted, so function(1) can always find a MRRT for a retry. But the source of problem is still that function (3) changes the parameter *cacheItem*)

RPangrle: As a general design goal in ADAL we're trying to avoid asynchronously calling backwards in the flow to steps we've already passed, because as you've discovered it's infinite recursive loop bait. In this case we'd probably be better off having it only retry the code it really needs to, which is probably that requestWithServer: call, not the whole validatedAcquireTokenByRefreshToken: call.

jasoncoolmax: Agree :)
Now the retry behaviour has been implemented at the web request level rather than the authentication request level.
But this line for fixing the infinite-retry bug is still needed, otherwise infinite retry still happens :P

RPangrle: It just means you found even more bugs! :)

The logic in this bit of code is a tad convoluted and the error handling is nowhere near as explicit as it should be. I would recommend instead checking for a AUTH_NON_PROTOCOL_ERROR in the response dictionary and sending that along to the completionBlock instead of executing any of the code in this block.

ADALiOS/ADALiOSTests/ADAuthenticationContextTests.m
```diff
@@ -1154,6 +1154,41 @@ -(void) testBadRefreshToken
     ADAssertLongEquals(0, [self cacheCount]);
 }
 
+-(void) testUnusualHttpResponse
```
RPangrle: :+1:

ADALiOS/ADALiOS/ADAuthenticationRequest+AcquireToken.m
jasoncoolmax: Sure, I will do it :)

ADALiOS/ADALiOS/ADAuthenticationRequest+AcquireAssertion.m
```diff
@@ -145,6 +145,7 @@ - (void)attemptToUseCacheItem:(ADTokenCacheItem*)item
     }
     
     //Now attempt to use the refresh token of the passed cache item:
+    BOOL isMultiresourceRefreshToken = item.multiResourceRefreshToken;
```
jasoncoolmax: Before item is passed as an parameter to an async call, we mark down whether it is a MRRT.
Such that we don't need to depend on the assumption that *item* cannot be modified.

ADALiOS/ADALiOS/ADAuthenticationRequest+WebRequest.m
```diff
@@ -234,6 +237,29 @@ - (void)requestWithServer:(NSString *)authorizationServer
                     }
                 }
                     break;
+                case 500:
+                case 503:
```
jasoncoolmax: Now the retry behaviour all happens at the web request level. Authentication request does not know about this.

RPangrle: :+1:

ADALiOS/ADALiOS/ADAuthenticationRequest+AcquireAssertion.m
RPangrle: There's something about this assumption that I find suspicious, but it's not your doing. Maybe in the future we'll add a test where we get back an RT after redeeming an MRRT, but I don't think that's a thing the server ever does to us.

RPangrle:
:shipit:

PR #439, opened 2015-12-16
3 logging function calls fixed to fit the new logging macro

jasoncoolmax:
There are 3 function calls not fixed to fit the new logging macro. Now fixed.


RPangrle:
:shipit:

Issue #380, opened 2015-09-15
Bundle is not correctly accessed in UIAlertView+Additions.m


cesarami:
The code in UIAlertView+Additions reads the strings from the main bundle, it should instead read the strings from the ADALiOS.bundle inside the main bundle.
UIAlertView does this:
```
[[UIAlertView alloc] initWithTitle:NSLocalizedString(@"Enter your credentials", nil) 
                                                    message:nil 
                                                    delegate:nil 
                                           cancelButtonTitle:NSLocalizedString(@"Cancel", nil) 
                                           otherButtonTitles: nil]; 
```
ADAuthenticationBroker also access the bundle but it does it correctly.

RandalliLama:
@cesarami In order to give some data on the severity priority of this issue can you tell me how this bug manifests itself at runtime.  What's the symptom of the bug.

cesarami:
I have not seen a bug from this yet. I asked Kanishk for an account that would cause this code path to be executed but none were available. 

In comparison, ADAuthenticationBroker has a frameworkBundle method
```
// Retrive the bundle containing the resources for the library. May return nil, if the bundle
// cannot be loaded.
+ (NSBundle *)frameworkBundle
{
    static NSBundle       *bundle     = nil;
    static dispatch_once_t predicate;
    
    @synchronized(self)
    {
        dispatch_once( &predicate,
                      ^{
                          NSString* mainBundlePath      = [[NSBundle mainBundle] resourcePath];
                          AD_LOG_VERBOSE_F(@"Resources Loading", @"Attempting to load resources from: %@", mainBundlePath);
                          
                          NSString* frameworkBundlePath = [mainBundlePath stringByAppendingPathComponent:@"ADALiOS.bundle"];
                          bundle = [NSBundle bundleWithPath:frameworkBundlePath];
                          if (!bundle)
                          {
                              AD_LOG_INFO_F(@"Resource Loading", @"Failed to load framework bundle. Application main bundle will be attempted.");
                          }
                      });
    }
    
    return bundle;
}
```
This looks like the correct way to do it, since you are loading the ADAL bundle inside the app's main bundle.

Issue #365, opened 2015-08-25
Revove kpanwar's dev account from test app config.


RandalliLama:
There shouldn't be any developer specific config in the base repo.

RPangrle:
Remove the DevelopmentTeam property as well.

PR #446, opened 2016-01-05
Changed to access framework bundle for NSLocalizedString

jasoncoolmax:
The change is mainly for
1. Changed to access framework bundle for NSLocalizedString (for issue #380); 
2. Moved framework specific utils out of ADAuthenticationBroker to a new class "ADALFrameworkUtils" (for issue #448);

Besides, I also fixed the following:
2. Removed developer team property of test app (for issue #365); 
3. Fixed three places where correlation id is not properly passed as parameter.


RPangrle:
:shipit:

Issue #443, opened 2015-12-28
NSString+ADHelperMethods does not parse unpadded Base64 data


mstritt-ka:
NSData in iOS does not properly handle unpadded Base64 data, as explained in:
http://stackoverflow.com/questions/21406482/nsdata-wont-accept-valid-base64-encoded-string

However the NSString+ADHelperMethods call, `adBase64UrlDecode`, depends on the NSData implementation. I received unpadded Base64 data for the "profile_info" in the authentication response, which was then not parsed. While it did not seem to affect the login, it can be easily fixed by either:
- adding padding characters client side before calling `NSData initWithBase64EncodedString`
OR
- properly padding the Base64 data server side before sending it to the client

*Note: using 'convergence-dev' branch.*
*Note: error received is:*
`
2015-12-28 12:22:49.702 <APP>[6258:5740506] ADALiOS [2015-12-28 20:22:49 - 6BE0CA80-1CF7-4A20-9D0D-4D5DE2D240A9] ERROR: Error raised: 15. Additional Information: Domain: ADAuthenticationErrorDomain Details: The profile_info contents cannot be parsed, ![NSString adIsStringNilOrBlank:decoded]. (profile_info = <Base64 data>). ErrorCode: 15.
`

RPangrle:
My attempts to slim down these helpers didn't work. @jasoncoolmax, can you restore the versions we have on master please?

PR #454, opened 2016-01-06
Revert back to our own Base64 encoding implementations

RPangrle:
Inaccurate method names led to confusion which led to switching things to Apple's Bas64 encoding utils when they should not have been. Furthermore other issues with padding on Apple's base64 encoding utils causes other issues. (Fix for #443)


jasoncoolmax:
:shipit:

Issue #438, opened 2015-12-15
Add service error info in ADAL acquireToken call back


liwa:
Can you add service info (error, correlation Id, etc) in acquireToken callback?
We intend to log these in some telemetry data and log data so that we have better understanding of how many auth failure are related to service side.
Especially, we need to identify the situation that the user cancel the request because the user sees auth failure and canceling is the only choice. 
Thanks,
Lin

liwa:
We want those error details in sign in flow and refresh token flow.

RPangrle:
@jasoncoolmax: I would start by adding correlation ID to the ADAuthenticationResult. The service error is already included in the authentication error (it's the "protocol code" bit), that of course is assuming that we actually hit the service, which isn't the case when tokens are removed underneath us.

PR #457, opened 2016-01-08
Added correlation id to ADAuthenticationResult

jasoncoolmax:
(for #438)

1. A property correlationId is added to ADAuthenticationResult. 
2. Overloaded init functions of ADAuthenticationResult are also added with one extra parameter, i.e. correlationId.
3. Correlation ID is passed for ADAuthenticaticationResult init function calls when correlation id is available.


RPangrle:
:shipit:

Issue #458, opened 2016-01-09
Remove 'withError:' from keychain APIs


RPangrle:
This implies that the error parameter has an effect on the behavior of the method, it does not.

Issue #442, opened 2015-12-19
Remove the -[ADTokenCacheStoring removeAllWithError:] API


RPangrle:
This API is way too easily abused, especially in a shared keychain environment, and has been causing us headaches.

eddiekim:
:+1: We've been burned by this in the past

Issue #392, opened 2015-10-02
Apple Bug: Intermittent keychain error -34018


siddheshcnaik:
Hi,

We are getting following error randomly. 
2015-10-01 17:30:02.159 [2151:542849] ADALiOS [2015-10-02 00:30:02 - E45569F5-D2A1-4B36-9FC0-6410BA2004D2] ERROR: Error raised: 11. Additional Information: Domain: ADAuthenticationErrorDomain ProtocolCode:(null) Details:Cannot read the items in the keychain. Error code: -34018. Query: {
   svce = "MSOpenTech.ADAL.1|aHR0cHM6Ly9sb2dpbi5taWNyb3NvZnRvbmxpbmUuY29tL2NvbW1vbg|aHR0cHM6Ly9vdXRsb29rLm9mZmljZTM2NS5jb20v|MGVjZjQ1MzItZjJmYS00YjZlLWI1OWUtM2EzYzBmOTc2ZWEx";
}. ErrorCode: 11.

This causes the Office 365 SDK to hang up.

Could you please help us with this.

Thanks

appcurry:
We are also running into this issue and it is causing our app to fail at random. The error does not get propagated to the top level call and it just hangs. 

Would appreciate any feedback from the developers of the Azure iOS SDK.

RPangrle:
This unfortunately is a known iOS bug: https://forums.developer.apple.com/thread/4743 It appears to happen more when you're attached to the debugger and when low memory warnings occur.

RPangrle:
The Office 365 SDK hanging is an issue for that SDK, not a specific ADAL issue. The error itself is an Apple bug, I don't see an ADAL bug here.

PR #459, opened 2016-01-10
Keychain Rewrite

RPangrle:
- There was a number of inefficiencies in the previous keychain implementation. It would call back into keychain many more times then were necessary to retrieve data it already had. Removing these inefficiencies allows for much more streamlined keychain code while still being entirely backwards compatible.
- Turn off ARC for Keychain Code. This is rumored to help deal with the -34018 problem, although it's not really a fix. (#392)
- Remove 'withError' from our keychain storage APIs. (#458)
- Move 'removeAll' to internal-test-tool-only status. (#442)


jasoncoolmax:
:shipit:

PR #460, opened 2016-01-11
We should not call -canOpenURL when checking redirect URI validity as…

RPangrle:
… that requires an app to put its own scheme on LSApplicationQuerySchemes, which strikes me as a bit much, if it's in the info.plist file that should be sufficient for proving that it's call-able.


jasoncoolmax:
:shipit:

PR #462, opened 2016-01-12
Support for building pod as a framework

nick-m42:
Add support for accessing resources in bundle when pod is built as a framework.

When the cocoapod is built as a framework (`use_frameworks!`) the bundle path for resources changes. This patch finds the new path if its a framework. Can the CocoaPods version also be bumped so this will be pulled by OneDriveSDK?

Thanks.


RPangrle:
We're not taking any more changes for 1.2.x. We're focussing on the 2.1 release that should be happening in the near future. That work is being done in the broker-dev branch. If you would like I can make that change there.

nick-m42:
Yes, please make the change there as it is needed to allow the pod to find the storyboard when built as a framework. Thanks!

PR #465, opened 2016-01-13
Copy changes from #462 over to broker-dev.

RPangrle:
From @nick-m42's PR (#462)

> Add support for accessing resources in bundle when pod is built as a framework.
>
> When the cocoapod is built as a framework (use_frameworks!) the bundle path for resources changes. This patch finds the new path if its a framework.


jasoncoolmax:
:shipit:

Issue #445, opened 2016-01-04
Tombstone RTs instead of Removing them on server refusal


RPangrle:
We want to be able to track the reason why the server rejected the RT, even if it was done in another app. To do so we're going to "tombstone" a RT by marking it as "dead" and adding to the keychain items properties for the correlation ID of the request that we got the error from and the bundle ID of the app where it happened.

PR #467, opened 2016-01-14
Issues/broker dev/keychain store tomb stone

jasoncoolmax:
Referring #445 
Tombstone feature added to ADKeychainTokenCacheStore.Now if an item will be marked as dead and left in keychain rather than being deleted.

In particular, the following changes are made:

1) property *correlationId* is added to ADTokenCacheStoreItem. If the token is tombstoned, the value will be the correlation ID of the request that we got the error from; if not, it is the correlation ID of the request through which we get the token.

2) property *markAsDead* is added to ADTokenCacheStoreItem. It shows whether the token is tombstoned.

3) unit tests for tombstone feature is added.

### DIFF
--- a/ADALiOS/ADALiOS/ADAuthenticationContext+TokenCaching.m
+++ b/ADALiOS/ADALiOS/ADAuthenticationContext+TokenCaching.m
@@ -64,6 +64,7 @@
 - (ADTokenCacheStoreItem*)findCacheItemWithKey:(ADTokenCacheStoreKey*) key
                                         userId:(ADUserIdentifier*)userId
                                 useAccessToken:(BOOL*) useAccessToken
+                          requestCorrelationId:(NSUUID*) requestCorrelationId
                                          error:(ADAuthenticationError* __autoreleasing*) error
 {
     if (!key || !self.tokenCacheStore)
@@ -94,8 +95,8 @@
         }
         else
         {
-            //We have a cache item that cannot be used anymore, remove it from the cache:
-            [self.tokenCacheStore removeItemWithKey:key userId:userId.userId error:nil];
+            //We have a cache item that cannot be used anymore, mark it as dead in the cache:
+            [self.tokenCacheStore tombstoneItem:item requestCorrelationId:[requestCorrelationId UUIDString] error:nil];
         }
     }
     *useAccessToken = false;//No item with suitable access token exists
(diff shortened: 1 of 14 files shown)


ADALiOS/ADALiOS/ADTokenCacheStoring.h
```diff
@@ -66,4 +66,27 @@
                    userId:(NSString *)userId
                     error:(ADAuthenticationError * __autoreleasing *)error;
 
+/*! Mark a cache item as dead instead of removing it.
+ @param item: the item being marked as dead. Key will be extracted from the item.
+ @param userId: The user for which the item will be removed. Can be nil, in which case items for all users with
+ the specified key will be removed.
+ @param error: in case of an error, if this parameter is not nil, it will be filled with
+ the error details. */
+- (void)tombstoneItem:(ADTokenCacheStoreItem *)item
```
RPangrle: This is a public interface, so anything you add to it also becomes public. Tombstoning is more a private implementation detail that we should minimize the exposure of.

ADALiOS/ADALiOS/ADTokenCacheStoreItem.h
```diff
@@ -46,6 +46,13 @@
 
 @property (retain) ADUserInformation* userInformation;
 
+/*! Correlation ID of the request through which we get the token.
+ If the token is tombstoned, the value will be the correlation ID of the request that we got the error from.  */
+@property (retain) NSString* correlationId;
+
+/* Whether the token is tombstoned. */
+@property BOOL markAsDead;
```
RPangrle: This is a verb phrase, a property should be a noun, preferably just "tombstone."

ADALiOS/ADALiOS/ADAuthenticationContext+TokenCaching.m
```diff
@@ -94,8 +95,8 @@ - (ADTokenCacheStoreItem*)findCacheItemWithKey:(ADTokenCacheStoreKey*) key
         }
         else
         {
-            //We have a cache item that cannot be used anymore, remove it from the cache:
-            [self.tokenCacheStore removeItemWithKey:key userId:userId.userId error:nil];
+            //We have a cache item that cannot be used anymore, mark it as dead in the cache:
```
RPangrle: I wouldn't bother tombstoning expired ATs. There's no question why those are gone.

ADALiOS/ADALiOS/ADKeychainTokenCacheStore.m
```diff
@@ -258,12 +269,14 @@ - (ADTokenCacheStoreItem*)itemFromKeyhainAttributes:(NSDictionary*)attrs
     for (NSDictionary* attrs in items)
     {
         ADTokenCacheStoreItem* item = [self itemFromKeyhainAttributes:attrs];
-        if (!item)
+        if (item)
         {
-            continue;
+            BOOL validity =  fromGraveyard ? [item markAsDead] : ![item markAsDead];
```
RPangrle: Let higher layers of code decide whether or not they want to deal with tombstoned items.

ADALiOS/ADALiOS/ADTokenCacheStoring.h
jasoncoolmax: Oh, I see, it is a private implementation. I will change it :)

ADALiOS/ADALiOS/ADTokenCacheStoreItem.h
jasoncoolmax: Sure :)

RPangrle:
Also, make sure you replace the refresh token with something that makes it obvious it's a tombstone, ie. "<tombstone>" followed by a short description of why it's tombstoned (the server error we got when it was rejected, or manually removed by <insert bundle id here>) This way if there are ADAL apps around that haven't been updated even if they try to use it the auth request will still fail (albeit not as gracefully as if they updated)

jasoncoolmax:
This branch for tombstone implementation is too old to be merged.

Issue #461, opened 2016-01-12
No Keychain Storage Implementation for OS X


RPangrle:
On OS X Universal we attempted to provide a keychain implementation, but it ran into a number of issues when different applications tried to share the same cache. On iOS access to the keychain is much more controller, but on OS X that can result in a large number of security dialogs popping up at the user. It's beyond scope for ADAL to attempt to solve cross-app token sharing on OS X and there will probably be other ways to achieve SSO on OS X that won't involve sharing keychain stores.

Developers that wish to persist tokens will have to provide a storage interface using the protocol in #449.

We will want to make sure that the README and any sample code we provide calls out this limitation on OS X.

Issue #449, opened 2016-01-05
New cache storage interface


RPangrle:
Our current pluggable caching interface is too specific, and make the architecture a lot messier. Change the interface to just be serialize and deserialize on a client ID basis.

RPangrle:
This will replace the cache interface on ADAuthenticationContext:

@protocol ADCacheStorage <NSObject>
​
- (NSData*)retrieveStorage;
- (void)saveToStorage:(NSData*)data;
​
@end

Issue #248, opened 2015-01-17
Tests to ensure that aquireToken with AD_PROMPT_NEVER still succeed when the main thread is blocked


RPangrle:
It's possible something is blocking the main thread while one of these APIs is called:

-(void)  acquireTokenWithResource: (NSString*) resource
                         clientId: (NSString*) clientId
                      redirectUri: (NSURL*) redirectUri
                   promptBehavior: (ADPromptBehavior) promptBehavior
                           userId: (NSString*) userId
             extraQueryParameters: (NSString*) queryParams
                  completionBlock: (ADAuthenticationCallback) completionBlock;

This can result in an unfortunate circumstance where a resource loop occurs resulting in a deadlock. Adding tests for this to make sure the call still works while the main thread is being blocked. (Namely remove the part in the test code where it keeps the run loop spinning) would be helpful for finding these bugs and preventing future ones from being added.

kpanwar:
Yes, this method can potentially be called. I'll remove the main queue call.
-(void) internalAcquireTokenByRefreshToken: (NSString*) refreshToken
                                  clientId: (NSString*) clientId
                                  resource: (NSString*) resource
                                    userId: (NSString*) userId
                                 cacheItem: (ADTokenCacheStoreItem*) cacheItem
                         validateAuthority: (BOOL) validateAuthority
                             correlationId: correlationId
                           completionBlock: (ADAuthenticationCallback)completionBlock

kpanwar:
Only call that is made on main thread is the one that launches the UI.If you pass PROMPT.NEVER then it should always execute on the background thread.

RPangrle:
We should have tests on the Silent APIs where we don't keep the main run loop spinning to make sure nothing inadvertently ends up trying to jump to the main thread that shouldn't.

PR #469, opened 2016-01-25
Keychain and related API Changes

RPangrle:
#461 #449 #470 #248

API Changes:
- Rename ADTokenCacheStore --> ADTokenCache
- Rename ADTokenCacheStoreItem -> ADTokenCacheItem
- Rename ADKeychainCacheStore -> ADKeychainCache
- Make ADKeychainCacheStore public.
- Remove the ADTokenCacheStoring interface
- Remove the ADTokenCacheStoring related params from ADAuthenticationContext
- Add params in ADAuthenticationContext init methods to allow a Keychain Sharing Group to be specified instead. (iOS Only)
- New private-only ADTokenCacheAccessor used internally for ADAuthenticationContext to be able to decouple the acquire token code from the underlying implementation
- New ADTokenCache class (public on Mac, private/test-only on iOS) to allow developers to persist tokens while still retaining ADAL control of the token cache schema
- Removed defaultTokenCacheStore from ADAuthenticationSettings
- ADKeychainTokenCacheStore will now use defaultKeychainGroup in its default initializer
- [ADAuthenticationSettings defaultKeychainGroup] now is prepopulated with "com.microsoft.adalcache"
- ADKeychainTokenCache will use the main bundle's identifier if no shared group is specified

Test Changes:
- ADAuthenticationContextTests have been completely rewritten.
- Any tests specific to acquiring tokens were put in ADAcquireTokenTests
- ADTestAuthenticationContext is removed
- All ADAcquireTokenTests use ADTestURLConnection (NSURLConnection mock) now
- ADAcquireTokenTests use a purely in-memory ADTokenCache for all tests 
- ADDefaultTokenCacheStoreTests were renamed ADKeychainTokenCacheTests
- Time consuming and repetitive tests were removed


ADALiOS/ADALiOS.xcodeproj/project.pbxproj
```diff
@@ -461,18 +465,14 @@
...
-				8B6113FD1868C59E007759C2 /* ADTestAuthenticationContext.m */,
-				8BB8346A1807BE3F007F9F0D /* ADAuthenticationContextTests.m */,
+				8BB8346A1807BE3F007F9F0D /* ADAcquireTokenTests.m */,
+				9430C3551C55862A00D6506D /* ADAuthorityValidationTests.m */,
+				9430C34D1C54320400D6506D /* ADAuthenticationContextTests.m */,
 				8BB8346C1807C5F5007F9F0D /* ADAuthenticationParametersTests.m */,
 				D6C38DC91BA0E74F00D62ABF /* ADBrokerKeyHelperTests.m */,
 				8B0965C417F25770002BDFB8 /* Supporting Files */,
-				8B598960181710FD00744AEE /* ADDefaultTokenCacheStoreTests.m */,
+				8B598960181710FD00744AEE /* ADKeychainTokenCacheStoreTests.m */,
```
jasoncoolmax: Given that "ADKeychainCacheStore" being changed to"ADKeychainCache", shouldn't this be "ADKeychainCacheTests"? :P

ADALiOS/ADALiOS/ADKeychainTokenCache.m
```diff
@@ -274,57 +225,77 @@ - (ADTokenCacheStoreItem*)itemFromKeyhainAttributes:(NSDictionary*)attrs
 #pragma mark -
 #pragma mark ADTokenCacheStoring implementation
```
jasoncoolmax: We don't have ADTokenCacheStoring anymore, no? :P

ADALiOS/ADALiOS/ADTokenCache.m
```diff
@@ -0,0 +1,546 @@
...
+// THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION
+// ANY IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A
+// PARTICULAR PURPOSE, MERCHANTABILITY OR NON-INFRINGEMENT.
+//
+// See the Apache License, Version 2.0 for the specific language
+// governing permissions and limitations under the License.
+
+//
+//  This class provides a ADTokenCacheStoring interface around the provided ADCacheStorage interface.
```
jasoncoolmax: No more ADTokenCacheStoring :)

jasoncoolmax:
I signed off :)

ADALiOS/ADALiOS/ADAL.h
```diff
@@ -19,12 +19,10 @@
 
 #import "ADLogger.h"
 #import "ADAuthenticationContext.h"
-#import "ADTokenCacheStoring.h"
 #import "ADAuthenticationError.h"
 #import "ADAuthenticationResult.h"
-#import "ADTokenCacheStoreItem.h"
+#import "ADTokenCacheItem.h"
 #import "ADUserInformation.h"
-#import "ADTokenCacheStoreKey.h"
```
kpanwar: Cache key is public interface supported by both .NET and Android. We should not remove it.

ADALiOS/ADALiOS/ADAuthenticationContext+TokenCaching.m
```diff
@@ -36,21 +37,12 @@ - (ADTokenCacheStoreItem*)extractCacheItemWithKey:(ADTokenCacheStoreKey*)key
...
-        }
-        
-        if (item && item.userInformation)
-        {
-            item = nil;//Different user id, just clear.
-        }
+    {
+        //ADFS fix, where the userId is not received by the server, but can be passed to the API:
+        //We didn't find element with the userId, try finding an item with a blank userId:
+        item = [self.tokenCacheStore getItemWithKey:key userId:@"" error:&localError];
```
kpanwar: why do you we need to pass an empty userId value here?

ADALiOS/ADALiOS/ADAuthenticationContext+TokenCaching.m
```diff
@@ -36,21 +37,12 @@ - (ADTokenCacheStoreItem*)extractCacheItemWithKey:(ADTokenCacheStoreKey*)key
     }
     
     ADAuthenticationError* localError;
-    ADTokenCacheStoreItem* item = [self.tokenCacheStore getItemWithKey:key userId:userId.userId error:&localError];
+    ADTokenCacheItem* item = [self.tokenCacheStore getItemWithKey:key userId:userId.userId error:&localError];
     if (!item && !localError && userId)
-    {//ADFS fix, where the userId is not received by the server, but can be passed to the API:
-        //We didn't find element with the userId, try finding an item with nil userId:
-        NSArray* items = [self.tokenCacheStore getItemsWithKey:key error:&localError];
-        if(items.count) {
```
kpanwar: this code is still required to handle ADFS 2.0 as an authority.

ADALiOS/ADALiOS/ADAuthenticationContext+TokenCaching.m
```diff
@@ -36,21 +37,12 @@ - (ADTokenCacheStoreItem*)extractCacheItemWithKey:(ADTokenCacheStoreKey*)key
...
-            item = items.firstObject;
-        }else{
-            item = nil;
-        }
-        
-        if (item && item.userInformation)
```
kpanwar: this checks if the user returned was from AAD/ADFS Blue, then ignore.

ADALiOS/ADALiOS/ADAL.h
RPangrle: What's it being used for though. The public API on the cache now is:

allItems and removeItem:, taking in a ADTokenCachItem, this matches the API surface on .NET. There is no public usage of ADTokenCacheKey w.r.t. caching anymore, as such it becomes an internal implementation detail and shouldn't be made public

ADALiOS/ADALiOS/ADAuthenticationContext.m
```diff
@@ -121,24 +127,36 @@ - (id)init
...
-        _authority = extractedAuthority;
-        _validateAuthority = bValidate;
-        _tokenCacheStore = tokenCache;
-        _credentialsType = AD_CREDENTIALS_EMBEDDED;
+        return nil;
     }
+    
+    _authority = extractedAuthority;
+    _validateAuthority = bValidate;
+    _credentialsType = AD_CREDENTIALS_EMBEDDED;;
```
kpanwar: consider renaming AD_CREDENTIAL_* to something meaningful? I like embedded vs Auto part.

ADALiOS/ADALiOS/ADInstanceDiscovery.m
```diff
@@ -164,16 +164,12 @@ -(void) validateAuthority: (NSString*) authority
...
-    
-    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_HIGH, 0), ^
-                   {
-                       //Nothing in the cache, ask the server:
-                       [self requestValidationOfAuthority:authority
-                                                     host:authorityHost
-                                         trustedAuthority:sTrustedAuthority
-                                            correlationId:correlationId
-                                          completionBlock:completionBlock];
-                   });
```
kpanwar: Office needed this to happen on a background thread instead of the main thread. thats why we added dispatch_async.

ADALiOS/ADALiOS/ADKeychainTokenCache.h
```diff
@@ -14,35 +14,33 @@
 // PARTICULAR PURPOSE, MERCHANTABILITY OR NON-INFRINGEMENT.
 //
 // See the Apache License, Version 2.0 for the specific language
-// governing permissions and limitations under the License.
-
-#import "ADTokenCacheStoring.h"
+// governing permissions and limitations under the License
```
kpanwar: not sure if a missing period violates license definition.

ADALiOS/ADALiOS/ADAuthenticationContext+TokenCaching.m
RPangrle: It's still there. I'm taking advantage of the fact that if we don't have an idtoken for an item we cache it with an empty string (@"") as the userid so we can retrieve only the item we care about.

ADALiOS/ADALiOS/ADAuthenticationContext+TokenCaching.m
RPangrle: See above.

ADALiOS/ADALiOS/ADInstanceDiscovery.m
RPangrle: That's an async call. It shouldn't matter.

ADALiOS/ADALiOS/ADKeychainTokenCache.m
```diff
@@ -387,35 +421,29 @@ - (void)addOrUpdateItem:(ADTokenCacheStoreItem *)item
...
-        if (error)
+        NSMutableDictionary* query = [self queryDictionaryForKey:nil userId:nil additional:nil];
+        OSStatus status = SecItemDelete((CFDictionaryRef)query);
+        [ADKeychainTokenCache checkStatus:status details:@"Failed to remove all" error:error];
+        
+        NSArray* items = [self allItems:nil];
+        if ([items count])
         {
-            *error = adError;
+            NSLog(@"!!!!!!!!!!!!!!!!!!!! %lu items remaining...", (unsigned long)items.count);
```
kpanwar: this should be logged via logger and not NSLog.

ADALiOS/ADALiOS/ADTokenCache.h
```diff
@@ -0,0 +1,53 @@
...
+// ANY IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A
+// PARTICULAR PURPOSE, MERCHANTABILITY OR NON-INFRINGEMENT.
+//
+// See the Apache License, Version 2.0 for the specific language
+// governing permissions and limitations under the License.
+
+
+#import <Foundation/Foundation.h>
+
+#define CURRENT_WRAPPER_CACHE_VERSION 1.0
```
kpanwar: what would be a case for wrapper version to be 1.1 vs 2.0? If there is no a big difference then consider using 1, 2, 3 etc..

ADALiOS/ADALiOS/ADTokenCache.m
```diff
@@ -0,0 +1,546 @@
...
+        return nil;
+    }
+}
+
+
+- (BOOL)deserialize:(nullable NSData*)data
+              error:(ADAuthenticationError * __nullable __autoreleasing * __nullable)error
+{
+    // Start by dropping our existing cache
+    _cache = nil;
```
kpanwar: failure to deserialize should not result in cache state change. Last state of the cache should be preserved. so, override cache only if deserialize was successful.

ADALiOS/ADALiOS/ADTokenCache.m
```diff
@@ -0,0 +1,546 @@
...
+    
+    [userTokens removeObjectForKey:key];
+    
+    // Check to see if we need to remove the idtoken
+    if (!userTokens.count)
+    {
+        [tokens removeObjectForKey:userId];
+        [[_cache objectForKey:@"idtokens"] removeObjectForKey:userId];
+    }
+
```
kpanwar: Consider adding logging statements for count and cache key. It will help make the logs read like functional specs.

ADALiOS/ADALiOS/ADTokenCache.m
```diff
@@ -0,0 +1,546 @@
...
+    NSDictionary* tokens = [cache objectForKey:@"tokens"];
+    
+    if (tokens)
+    {
+        CHECK_ERROR([tokens isKindOfClass:[NSMutableDictionary class]], AD_ERROR_BAD_CACHE_FORMAT, @"tokens must be a mutable dictionary.");
+        for (id userId in tokens)
+        {
+            // On the second level we're expecting NSDictionaries keyed off of the user ids (an NSString*) or
+            // NSNull if no userId was available.
+            CHECK_ERROR([userId isKindOfClass:[NSString class]] || [userId isKindOfClass:[NSNull class]], AD_ERROR_BAD_CACHE_FORMAT, @"User ID key not the expected class type");
```
kpanwar: Not sure is using NSNull for userId as a key would work for multiple accounts. You may not get a user id back in case of federation with Ping/ADFS.

ADALiOS/ADALiOS/ADTokenCache.m
```diff
@@ -0,0 +1,546 @@
...
+    if (items.count == 1)
+    {
+        return items.firstObject;
+    }
+    
+    ADAuthenticationError* adError =
+    [ADAuthenticationError errorFromAuthenticationError:AD_ERROR_MULTIPLE_USERS
+                                           protocolCode:nil
+                                           errorDetails:@"The token cache store for this resource contain more than one user. Please set the 'userId' parameter to determine which one to be used."];
+    if (error)
```
kpanwar: shouldn't if(error) be checked right after calling getItemsWithKey?

ADALiOS/ADALiOS/ADTokenCacheItem.m
```diff
@@ -169,4 +186,48 @@ - (NSString*)description
...
+
+- (ADUserInformation *)userInformation
+{
+    return _userInformation;
+}
+
+- (void)setUserInformation:(ADUserInformation *)userInformation
+{
+    _userInformation = userInformation;
+    [self calculateHash];
```
kpanwar: there are higher chances of introducing a bug if a new property is added and developer forgets to call calculateHash. Consider calling this function before returning the hash. So you would have getHashcode() or something like that.

RPangrle: calculateHash is expensive and -hash gets called a lot when using Foundation Collections. -hash is supposed to be a fast call.

ADALiOS/ADALiOS/ADKeychainTokenCache.m
RPangrle: It's a test utility. The unit tests squash some of the logging and I wanted to make sure it got through.

ADALiOS/ADALiOS/ADTokenCache.m
RPangrle: I was using NSNull for the user id key when I didn't get a user ID as I can't add things to a dictionary for a nil key. I ended up switching to an empty string (@"") to match the behavior of the keychain cache. I can remove this NSNull check here though.

ADALiOS/ADALiOS/ADTokenCache.m
RPangrle: This if (error) isn't checking the getItemsWithKey call. That call is being checked with if (!items) right after the call.

This if (error) is checking to see if the developer supplied the optional error parameter to make sure I don't segfault if it wasn't provided.

ADALiOS/ADALiOS/ADTokenCache.h
RPangrle: I was thinking 2.x would be convergence cache, and 1.x would be pre-convergence. I'm not sure the right answer there.

ADALiOS/ADALiOS/ADTokenCache.m
RPangrle: Makes sense.

ADALiOS/ADALiOS/ADTokenCache.m
RPangrle: Filed as #471

ADALiOS/ADALiOS/ADKeychainTokenCache.h
RPangrle: +1

ADALiOS/ADALiOS/ADTokenCache.m
RPangrle: :+1:

ADALiOS/ADALiOS/ADTokenCacheItem.m
kpanwar: How expensive exactly? Hash is something we introduced. It is not a system property override as far as I can tell.

ADALiOS/ADALiOS/ADTokenCacheItem.m
```diff
@@ -17,16 +17,33 @@
...
 @synthesize multiResourceRefreshToken;
 
+- (NSUInteger)hash
+{
+    return _hash;
+}
+
+- (void)calculateHash
+{
+    _hash = [[NSString stringWithFormat:@"%@%@%@%@", _resource, _authority, _clientId, _userInformation.userId] hash];
```
kpanwar: values are ## separated.

ADALiOS/ADALiOS/ADTokenCache.h
kpanwar: Keep it simple with non floating point versions :)

ADALiOS/ADALiOS/ADTokenCacheItem.m
RPangrle: -hash comes from here: https://developer.apple.com/library/ios/documentation/Cocoa/Reference/Foundation/Protocols/NSObject_Protocol/

ADALiOS/ADALiOS/ADTokenCacheItem.m
RPangrle: It doesn't matter. This doesn't need to duplicate ADTokenCacheKey (if you notice, there are more fields here...) this is trying to pull out an integer value that can help sort the object, by definition it's not going to have the same resolution as the string that was hashed.

ADALiOS/ADALiOS/ADKeychainTokenCache.m
RandalliLama: Would it help to add a comment to that effect?

Issue #464, opened 2016-01-13
Move public headers into a subfolder


RPangrle:
It's extremely unobvious right now which headers are public vs private, we should fix that.

Issue #456, opened 2016-01-08
iPAL Clean Up


RPangrle:
A lot of the "iPAL" isn't platform-specific from a Mac vs iOS perspective. This should be cleaned up to make explicit what is truly mac or iOS only.

Issue #455, opened 2016-01-08
Remove "iOS" from project and file names


RPangrle:
Part of making ADAL truly work across Mac and iOS should be not requiring ifdefs just for the #imports. Change file names so that they don't mention "iOS" or "Mac" unless the file is truly platform specific.

Issue #373, opened 2015-09-04
ADErrorCodes.h is marked as a "Project" header not a "Public" header on OSX


HiKumar-MS:
ADAL-OSX project marks ADErrorCodes.h as a "Project" header, when it should be a "Public" header to allow developers to use the error code enums to better handle the errors returned by ADAL library. This header is public on iOS, and appears to be an oversight.

PR #472, opened 2016-01-27
Rearrange the ADAL project structure

RPangrle:
#464 #456 #455 #373 #353 (?) 

In preparation for impending Mac support "iOS" has been removed from the project and folder names, and the folder structure has been cleaned up to group source files as logically as possible and make it easier to tell what files are Mac or iOS specific or a public header. The general file structure now looks like:

```
ADAL/
	src/				Top level directory for all production ADAL source files
		public/			All ADAL public headers now have to be in the public directory
		broker/			Broker invocation and utility related files
		cache/			Token caching related files
		request/		ADAuthenticationRequest and ADWebRequest
		ui/				UI related files (popping up the web view)
		urlprotocol/	NSURLProtocol implementation and helper (NTLM, etc.) 
		utils/			General utilities (ie. categories on system methods)
		workplacejoin/	Workplace Join related utilities
	tests/				Top level directory for test files
	resources/			info.plist files, storyboards, etc.

```

At any level in this structure there can also be ios and mac subfolders for any iOS or Mac only implementation files. Anything that is not in a platform specific subfolder should be buildable on both platforms, although potentially with ifdefs for minor differences. Major implementation differences should be in separate files.

I also removed some files that aren't used at all, and made ADInstanceDiscovery private, as a similar interface isn't public on other platforms and it really should be just an implementation detail.


RPangrle:
@jasoncoolmax you'll want to look over this.

jasoncoolmax:
I am on it :P

jasoncoolmax:
The folder structure is now way way better :+1: 
I signed off :)

Issue #290, opened 2015-03-30
[OSX] Module definition


damienpontifex:
With Xcode now supporting modules (and Swift's requirement for modules) can the OS X project be updated with build settings for DEFINES_MODULES to Yes?

I believe the target name will also need renaming if this occurs to conform with c99 standards i.e. rename to ADAL_OSX vs the hyphen in the current name.

RPangrle:
@damienpontifex Can you take a peek at the broker-dev branch and see if this fixes your issue?

damienpontifex:
Initial test still getting the error 'No such module ADAL'. Still trying to debug what is missing. I wouldn't have thought a custom module map was required, the Xcode framework target should be able to generate everything required.

One thing to note is the 'ADAL Mac' framework target doesn't seem to have sufficient headers marked as Public.

On another note looking at #473 about mac & iOS support, why not use `WKWebView` which is apple's modern web rendering view and exists on both platforms?

RPangrle:
Give broker-mac-ui-impl a shot then. I added all the public headers into the framework and the module map. 

The main reason to not use WKWebView is we haven't in the past, already have the the code for UIWebView/WebView and there's not much to gain from using it right now as we're not terribly JavaScript heavy. Also I might still have to support iOS 7. :-/ Not entirely sure on that last point. I could see once iOS 10 comes out making the switch.

damienpontifex:
That seemed to work on topics/broker-mac-ui-impl branch :)

I did change the 'Skip Install' build setting to 'No' so I could archive the product and get the release output.

Thanks for this, feel free to close this issue.

PR #473, opened 2016-01-27
Mac Support: Round 1

RPangrle:
#373 #452 #290 

This is the first round of Mac support. It doesn't represent the finished product, but gets us at a starting point where everything is building and we can have people working on top of this without having to fork for long periods of time and try to merge some potentially large architectural changes over.

Most iOS specific code is protected by ifdefs. The larger ifdefs should get moved off to platform specific files and utilities, although some of the 1-5 line ifdefs will probably be left in place.

Some iOS only features are here and compiling on Mac, although with access to those code paths blocked. ex. Broker and Workplace Join support. In both cases a check to use those features will return NO. In both cases we need to define what those features should even look like on Mac OS X, and some of the crypto in pkeyauth will need mac specific code as the crypto calls we use on iOS aren't available on OS X. Keeping this code compiling on Mac at least ensures that we don't accidentally go backwards.


RPangrle:
@kpanwar

jasoncoolmax:
I am on it :)

ADAL/src/public/ADAL.h
```diff
@@ -26,6 +26,21 @@ FOUNDATION_EXPORT double ADALFrameworkVersionNumber;
 //! Project version string for ADALiOSFramework.
 FOUNDATION_EXPORT const unsigned char ADALFrameworkVersionString[];
 
+#if TARGET_OS_IPHONE
+//iOS:
```
jasoncoolmax: Any reason why we use IPHONE to represent iOS? :P

RPangrle: We don't, Apple does. That's a system macro. It dates back to when iOS was called "iPhone OS"

RPangrle: Fun fact: TARGET_OS_MAC is defined as true on both iOS and Mac OS, which is why !TARGET_OS_IPHONE is used to determine if we're on a Mac or not.

jasoncoolmax: Got it, haha :)

jasoncoolmax:
I signed off :)

ADAL/src/public/ADAuthenticationSettings.h
```diff
@@ -41,6 +43,7 @@
 /*! The dispatch queue to be used for the asynchronous calls. */
 @property (retain) dispatch_queue_t dispatchQueue;
```
kpanwar: Lets just remove it.

ADAL/src/ui/mac/ADCredentialCollectionController.m
```diff
@@ -0,0 +1,44 @@
...
+@synthesize usernameLabel = _usernameLabel;
+@synthesize passwordLabel = _passwordLabel;
+
+-(id) init
+{
+    self = [super init];
+    if(self)
+    {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
```
kpanwar: ignore warnings?!?! Can use initWithNibNamed?

kpanwar:
Sign-off with comments,

ADAL/src/ui/mac/ADCredentialCollectionController.m
RPangrle: #474

ADAL/src/public/ADAuthenticationSettings.h
RPangrle: #475

Issue #476, opened 2016-01-28
Remove Storyboards, replace with programatic UI


RPangrle:
The only resources we have are the storyboards, and they're a bit of a pain to manage and the code to use them is a tad convoluted. Building the UI programmatically is easier and cleaner, and lets us get rid of the ADALiOSBundle, making setting up an ADAL app that much easier!

PR #477, opened 2016-01-29
UI Refactor. iOS UI Rewrite

RPangrle:
#452 

- Storyboards are gone! UI is now built programatically. This allowed me to simplify our UI code quite a bit. (#476)
- All business logic moved into ADWebAuthController (which is used on both Mac and iOS)
- UI code for creating and presenting the authentication view is now in ADAuthenticationViewController, which will have different implementations based on platform.
- ADAuthenticationViewController uses a delegate to pass up web view events (page loading, will load, etc) back to ADWebAuthController
- Removed ADAuthenticationWebViewController, we don't need multiple view controllers for such a simple scenario.


RPangrle:
@brentschmaltz you'll be interested in this code.

jasoncoolmax:
I signed off.

brentschmaltz:
@RPangrle I tired to use 'newbie' when asking about general stuff :-)

Issue #451, opened 2016-01-05
New keychain removal APIs to replace -removeAll


RPangrle:
-[ADTokenCache removeAllForClientId:] and-[ADTokenCache removeAllForIdentifier: clientId:] These two closely match the things the developer is most likely trying to actually do. Either remove all the tokens for their own app (clean wipe) or for a specific user (sign out). In both cases we'll tombstone the RTs so we will have more data if we're tracking down keychain stomping issues.

RPangrle:
We definitely should support these methods on ADKeychainCacheStore

PR #478, opened 2016-01-29
Issues/broker dev/tombstone wrapper

jasoncoolmax:
Referring #445 and #451 
- Tombstone feature added; 
- [ADKeychainTokenCache removeAllForIdentifier:] and [ADKeychainTokenCache removeAllForClientId:] added.

Details:
1) Property *tombstone*, which is a NSMutableDictionary, is added to ADTokenCacheItem. Such a property is nil for normal cache item, but not nil for a tombstone.
Property *tombstone* is supposed to record the following information if item is set as tombstone:
 @"bundleId":Bundle ID of the app which tombstones the token.
 @"correlationId":correlation ID of the request that we got the error from.
 @"protocolCode":error code returned by the server for the rejected RT
 @"errorDetails":error details of the rejected RT

2) Changes have been made for [ADKeychainTokenCache removeItem:]: if an item contains a refresh token, it will be set as a tombstone and updated in cache; if it does not, it will be directly deleted.

3)- [ADKeychainTokenCache allItems:] and - [ADKeychainTokenCache getItemWithKey:] have been updated to filter out tombstone items before return;

4) logging has been improved to make good use of tombstones in ADKeychainTokenCache. E.g., when [ADKeychainTokenCache getItemWithKey:] cannot retrieve anything but tombstone items, information recorded in the tombstone will be logged, such that we can better understand who is responsible for previously deleting the item by looking at the recorded bundle ID; why token is deleted by looking at the errorDetails.

5) unit tests for tombstone feature and removeAllForIdentifier/removeAllForClientId are added.


ADAL/src/cache/ADAuthenticationContext+TokenCaching.h
```diff
@@ -31,11 +31,13 @@
 //the item to be stored.
 - (void)updateCacheToResult:(ADAuthenticationResult*)result
                   cacheItem:(ADTokenCacheItem*)cacheItem
-           withRefreshToken:(NSString*)refreshToken;
+           withRefreshToken:(NSString*)refreshToken
+       requestCorrelationId:(NSUUID*)requestCorrelationId;
```
jasoncoolmax: Rejected refresh token is removed within this function, so request correlation ID should be passed in in order to do the tombstone action.

ADAL/src/cache/ADTokenCacheAccessor.h
```diff
@@ -20,6 +20,7 @@
 
 @class ADTokenCacheKey;
 @class ADTokenCacheItem;
+@class ADAuthenticationError;
```
jasoncoolmax: When I tried to include ADTokenCacheAccessor.h, I got some errors asking for forward declaration :P

ADAL/src/cache/ADTokenCacheItem.m
```diff
@@ -32,6 +32,19 @@ @implementation ADTokenCacheItem
     ADUserInformation* _userInformation;
 }
 
+- (id)init
```
jasoncoolmax: init function is added to initialize property *tombstone* and *bundleId*

ADAL/src/cache/ios/ADCacheTombstoneWrapper.m
```diff
@@ -0,0 +1,117 @@
...
+    ADTokenCacheItem* item = [_cache getItemWithKey:key userId:userId error:error];
+    if ([item tombstone])
+    {
+        return nil;
+    }
+    else
+    {
+        return item;
+    }
+}
```
jasoncoolmax: Many places in the project assume that getItemWithKey will give them a normal item rather than a tombstone, so I filter out the tombstones before return.

In other words, only function allItems can retrieve tombstones now. Not sure whether it is ok :P

RPangrle: Make sure you log here that you found a tombstone.

ADAL/src/cache/ADTokenCacheItem.m
```diff
@@ -32,6 +32,19 @@ @implementation ADTokenCacheItem
...
+{
+    if (!(self = [super init]))
+    {
+        return nil;
+    }
+    
+    [self setTombstone:NO];
```
RPangrle: default value for all properties on an ObjC object is 0/NO/nil

ADAL/src/cache/ios/ADCacheTombstoneWrapper.m
jasoncoolmax: Sure :)

ADAL/src/cache/ADTokenCacheItem.m
jasoncoolmax: Glad to know that :)

ADAL/src/cache/ios/ADCacheTombstoneWrapper.m
jasoncoolmax: I have added the log, but don't know why the above snippet is not updated :P
Please look at the File Change tab for the update :)

ADAL/src/cache/ADTokenCacheAccessor.h
RPangrle: That means you need to include those files as well.

ADAL/src/cache/ios/ADCacheTombstoneWrapper.m
```diff
@@ -0,0 +1,118 @@
...
+ @param  item    The item to remove/tombstone from the cache, depending on the tombstone property of item.
+ @param  error   (Optional) In the case of an error this will be filled with the
+ error details.
+ 
+ @return YES if the item was successfully removed/tombstoned. NO otherwise.
+ */
+- (BOOL)removeItem:(nonnull ADTokenCacheItem *)item
+             error:(ADAuthenticationError * __nullable __autoreleasing * __nullable)error
+{
+    if ([item tombstone])
```
RPangrle: Unfortunately this code now needs to be embedded in ADKeychainTokenCache, because that's the public interface apps will be able to play on :-/

jasoncoolmax: Sure :)

So developers can tombstone an item manually, right?

RPangrle: No, so developers don't get a choice in the matter. The point of this is to be able to track down misbehaving apps that are deleting things they shouldn't.

jasoncoolmax: Should we force developers to set the tombstone property to be YES before they call the [ADKeychainTokenCache removeItem]?

RPangrle: No, we should do that for them.

ADAL/src/cache/ADAuthenticationContext+TokenCaching.h
RPangrle: Hrm, I would have thought we'd already have the token cache item with the refresh token laying around somewhere. It's possible it's just not factored well enough. I'll flag this code path for some cleaning up in the future.

ADAL/src/cache/ADAuthenticationContext+TokenCaching.m
```diff
@@ -188,14 +191,16 @@ - (void)updateCacheToResult:(ADAuthenticationResult*)result
...
-            //The refresh token didn't work. We need to clear this refresh item from the cache.
+            //The refresh token didn't work. We need to tombstone this refresh item in the cache.
             ADTokenCacheKey* exactKey = [cacheItem extractKey:nil];
             if (exactKey)
             {
                 ADTokenCacheItem* existing = [tokenCacheStoreInstance getItemWithKey:exactKey userId:cacheItem.userInformation.userId error:nil];
                 if ([refreshToken isEqualToString:existing.refreshToken])//If still there, attempt to remove
                 {
-                    AD_LOG_VERBOSE_F(@"Token cache store", [self correlationId], @"Removing cache for resource: %@", cacheItem.resource);
+                    AD_LOG_VERBOSE_F(@"Token cache store", [self correlationId], @"Tombstoning cache for resource: %@", cacheItem.resource);
```
RPangrle: Use the correlationID passed into the method.

ADAL/src/cache/ADTokenCacheItem+Internal.m
```diff
@@ -197,6 +197,7 @@ - (BOOL)fillItemWithResponse:(NSDictionary*)responseDictionary
     FILL_FIELD(accessToken, OAUTH2_ACCESS_TOKEN);
     FILL_FIELD(refreshToken, OAUTH2_REFRESH_TOKEN);
     FILL_FIELD(accessTokenType, OAUTH2_TOKEN_TYPE);
+    FILL_FIELD(correlationId, OAUTH2_CORRELATION_ID_RESPONSE);
```
RPangrle: We only really care about the correlation ID when it is a tombstone.

ADAL/src/cache/ADTokenCacheItem.m
```diff
@@ -66,6 +79,9 @@ -(id) copyWithZone:(NSZone*) zone
     item.expiresOn = [self.expiresOn copyWithZone:zone];
     item.userInformation = [self.userInformation copyWithZone:zone];
     item.sessionKey = [self.sessionKey copyWithZone:zone];
+    item.tombstone = [self tombstone];
```
RPangrle: We should have the error details stashed away as well.

ADAL/src/cache/ios/ADKeychainTokenCache.m
```diff
@@ -245,30 +245,48 @@ - (ADTokenCacheItem*)itemFromKeyhainAttributes:(NSDictionary*)attrs
...
     if (!key)
     {
         return NO;
     }
-    
-    
     NSMutableDictionary* query = [self queryDictionaryForKey:key
                                                       userId:item.userInformation.userId
                                                   additional:nil];
     OSStatus status = SecItemDelete((CFDictionaryRef)query);
```
RPangrle: Just do a SecItemUpdate instead. If that fails then we know the item wasn't in the cache in the first place.

ADAL/src/cache/ADTokenCacheItem.m
```diff
@@ -141,6 +160,9 @@ -(id) initWithCoder:(NSCoder *)aDecoder
         self.refreshToken = [aDecoder decodeObjectOfClass:[NSString class] forKey:@"refreshToken"];
         self.expiresOn = [aDecoder decodeObjectOfClass:[NSDate class] forKey:@"expiresOn"];
         self.userInformation = [aDecoder decodeObjectOfClass:[ADUserInformation class] forKey:@"userInformation"];
+        self.tombstone = [aDecoder decodeBoolForKey:@"tombstone"];
```
RPangrle: Here's an idea, have tombstone be an NSDictionary instead and put all of the various fields we only care about if it's a tombstone (correlation ID, bundle ID, error details) in there. That'll reduce the risk on things getting lost between version changes.

jasoncoolmax: That's a good idea :)

ADAL/src/cache/ios/ADKeychainTokenCache.m
jasoncoolmax: Oh, yes, we should use SecItemUpdate :P

RPangrle: Actually, check if there's a RT in the item, if there isn't just do the SecItemDelete. If there is then do a SecItemUpdate with all the tombstone info.

RPangrle: As a general rule with keychain (and pretty much any database) the fewer discrete queries you can do something in the better.

ADAL/src/cache/ADAuthenticationContext+TokenCaching.m
jasoncoolmax: Sure :)

ADAL/src/cache/ADTokenCacheItem+Internal.m
jasoncoolmax: Got it :)

ADAL/src/cache/ADAuthenticationContext+TokenCaching.h
jasoncoolmax: OK :)

ADAL/src/cache/ADAuthenticationContext+TokenCaching.m
```diff
@@ -203,19 +209,23 @@ - (void)updateCacheToResult:(ADAuthenticationResult*)result
...
                 if (broadKey)
                 {
                     ADTokenCacheItem* broadItem = [tokenCacheStoreInstance getItemWithKey:broadKey userId:cacheItem.userInformation.userId error:nil];
                     if (broadItem && [refreshToken isEqualToString:broadItem.refreshToken])//Remove if still there
                     {
-                        AD_LOG_VERBOSE_F(@"Token cache store", [self correlationId], @"Removing multi-resource refresh token for authority: %@", self.authority);
+                        AD_LOG_VERBOSE_F(@"Token cache store", requestCorrelationId, @"Tombstoning multi-resource refresh token for authority: %@", self.authority);
+                        //set request correlation ID and error details in tombstone property
+                        [broadItem setTombstone:[NSMutableDictionary dictionaryWithDictionary:@{ @"correlationId" : [requestCorrelationId UUIDString],
+                                                                                                 @"errorDetails" : [result.error errorDetails]}]];
```
RPangrle: Can you add protocolCode as well too? Thanks.

jasoncoolmax: Yessir :)

ADAL/src/cache/ADAuthenticationContext+TokenCaching.m
```diff
@@ -188,14 +191,17 @@ - (void)updateCacheToResult:(ADAuthenticationResult*)result
...
             {
                 ADTokenCacheItem* existing = [tokenCacheStoreInstance getItemWithKey:exactKey userId:cacheItem.userInformation.userId error:nil];
                 if ([refreshToken isEqualToString:existing.refreshToken])//If still there, attempt to remove
                 {
-                    AD_LOG_VERBOSE_F(@"Token cache store", [self correlationId], @"Removing cache for resource: %@", cacheItem.resource);
+                    AD_LOG_VERBOSE_F(@"Token cache store", requestCorrelationId, @"Tombstoning cache for resource: %@", cacheItem.resource);
+                    //set request correlation ID and error details in tombstone property
+                    [existing setTombstone:[NSMutableDictionary dictionaryWithDictionary:@{ @"correlationId" : [requestCorrelationId UUIDString],
+                                                                                            @"errorDetails" : [result.error errorDetails]}]];
                     [tokenCacheStoreInstance removeItem:existing error:nil];
```
RPangrle: Call addOrUpdateItem: here instead. It'll (probably) confuse the underlying caching code less.

ADAL/src/cache/ios/ADKeychainTokenCache.m
```diff
@@ -245,32 +245,98 @@ - (ADTokenCacheItem*)itemFromKeyhainAttributes:(NSDictionary*)attrs
...
+        NSData* itemData = [NSKeyedArchiver archivedDataWithRootObject:tombstoneItem];
+        NSMutableDictionary * itemUpdate = [NSMutableDictionary dictionaryWithDictionary:@{ (id)kSecValueData : itemData,
+                                                                                              (id)kSecAttrAccessible : (id)kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly}];
+        [tombstoneItem release];
+        
+        //update the tombstone item to keychain
+        OSStatus status = SecItemUpdate((CFDictionaryRef)query, (CFDictionaryRef)itemUpdate);
+        return [ADKeychainTokenCache checkStatus:status details:@"Failed to remove item from keychain" error:error];
+    }
+    else
```
RPangrle: You can put this up at the top as a if ([NSString adIsStringNilOrBlank:item.refreshToken]) block and early return, then the rest of the code doesn't need to be indented!

ADAL/src/cache/ios/ADKeychainTokenCache.m
```diff
@@ -245,32 +245,98 @@ - (ADTokenCacheItem*)itemFromKeyhainAttributes:(NSDictionary*)attrs
...
+        //set the item to tombstone
+        ADTokenCacheItem* tombstoneItem = [item copy];
+        [self setCacheItemAsTombstone:tombstoneItem];
+        NSData* itemData = [NSKeyedArchiver archivedDataWithRootObject:tombstoneItem];
+        NSMutableDictionary * itemUpdate = [NSMutableDictionary dictionaryWithDictionary:@{ (id)kSecValueData : itemData,
+                                                                                              (id)kSecAttrAccessible : (id)kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly}];
+        [tombstoneItem release];
+        
+        //update the tombstone item to keychain
+        OSStatus status = SecItemUpdate((CFDictionaryRef)query, (CFDictionaryRef)itemUpdate);
```
RPangrle: Sorry, I gave you the wrong comment here, just reuse the code from addOrUpdateItem:, don't bother copying & pasting.

jasoncoolmax: Sure, no problem :)

ADAL/src/cache/ios/ADKeychainTokenCache.m
jasoncoolmax: Yeah, that will look nicer, i will do it.

ADAL/src/cache/ADAuthenticationContext+TokenCaching.m
jasoncoolmax: Agree :)

jasoncoolmax: I found a problem here, after changing *removeItem* to *addOrUpdateItem*, it does not work for Mac because ADTokenCache does not support tombstone (yet) and it expects *removeItem* being called and items being deleted :P

Any hint on how to proceed with it? :)

RPangrle: By problem you mean tests are failing? We should update the tests to expect the tombstones to be there might need to update tests to not count tombstones. We also could update allItems to not return tombstones. I'm fine with either.

jasoncoolmax: So we need to udate both -getItemWithKey: and -allItems: to filter out tombstones, right?

RPangrle: getItemWithKey: is already filtering out tombstones

ADAL/src/cache/ios/ADKeychainTokenCache.m
```diff
@@ -373,7 +439,16 @@ - (ADTokenCacheItem*)getItemWithKey:(ADTokenCacheKey *)key
...
     
-    return items.firstObject;
+    //filter out tombstone
+    if (![items.firstObject tombstone])
+    {
+        return items.firstObject;
+    }
+    else
+    {
+        [self logItem:items.firstObject message:@"Found a tombstone"];
```
RPangrle: Log out the tombstone details here. We don't need to log hashes of the fields because the refreshToken should just be "<tombstone>"

ADAL/src/cache/ADAuthenticationContext+TokenCaching.m
jasoncoolmax: I mean getItemWithKey: of ADTokenCache, not ADKeychainTokenCache :P

RPangrle: Probably :(

ADAL/src/cache/ios/ADKeychainTokenCache.m
```diff
@@ -245,30 +259,138 @@ - (ADTokenCacheItem*)itemFromKeyhainAttributes:(NSDictionary*)attrs
...
+            if (*error)
+            {
+                deleteSuccessful = NO;
+            }
+        }
+    }
+    return deleteSuccessful;
+}
+
+- (BOOL)removeAllForIdentifier:(NSString * __nonnull)userId
```
RPangrle: Change this to "ForUserId:" we use "Identifier" when we take in a ADUserIdentifier.

ADAL/src/cache/ios/ADKeychainTokenCache.m
```diff
@@ -245,30 +262,138 @@ - (ADTokenCacheItem*)itemFromKeyhainAttributes:(NSDictionary*)attrs
...
+        }
+    }
+    SAFE_ARC_AUTORELEASE(itemsKept);
+    return itemsKept;
+}
+
+- (BOOL)removeAllForClientId:(NSString * __nonnull)clientId
+                       error:(ADAuthenticationError * __nullable __autoreleasing * __nullable)error
+{
+    AD_LOG_VERBOSE_F(@"Keychain token cache store", nil, @"Removing items for client <%@>", clientId);
```
RPangrle: Make this a WARNING level log statement.

ADAL/src/cache/ios/ADKeychainTokenCache.m
```diff
@@ -245,30 +262,138 @@ - (ADTokenCacheItem*)itemFromKeyhainAttributes:(NSDictionary*)attrs
...
+        }
+    }
+    return deleteSuccessful;
+}
+
+- (BOOL)removeAllForUserId:(NSString * __nonnull)userId
+                      clientId:(NSString * __nonnull)clientId
+                         error:(ADAuthenticationError * __nullable __autoreleasing * __nullable)error
+{
+    AD_LOG_VERBOSE_F(@"Keychain token cache store", nil, @"Removing items for user <%@> + client <%@>", userId, clientId);
```
RPangrle: Make this a WARNING level log statement

RPangrle:
While you're at it make sure we have a log statement on removeItem:. Once you're done with these comments you're good to check in. :shipit:

Issue #453, opened 2016-01-06
Family of Client IDs Support


RPangrle:
Future server work will allow client IDs to be grouped on the server side in a way where a RT for one client ID can be redeemed for a AT and RT for a different client ID as long as they're in the same group. This will move us closer to being able to provide SSO-like functionality between apps without requiring the broker (or workplace join).

PR #485, opened 2016-02-02
Family of Client IDs

RPangrle:
#453 

Implementation of family-of-client-IDs as far as we understand it right now. ADAL_CLIENT_FAMILY_ID is a made up value for now until we get what the real key will be from the server guys.

This change caches the "family ID" that we get from the server, and when we try to acquire a token, and don't have an AT or RT for that client ID available we look through the cache for an item for that user tagged with a family ID.

### DIFF
--- a/ADAL/src/cache/ADAuthenticationContext+TokenCaching.h
+++ b/ADAL/src/cache/ADAuthenticationContext+TokenCaching.h
@@ -22,23 +22,26 @@
 
 //Checks the cache for item that can be used to get directly or indirectly an access token.
 //Checks the multi-resource refresh tokens too.
-- (ADTokenCacheItem*)findCacheItemWithKey:(ADTokenCacheKey*) key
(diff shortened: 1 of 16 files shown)


ADAL/src/cache/ADAuthenticationContext+TokenCaching.h
```diff
@@ -22,23 +22,26 @@
 
 //Checks the cache for item that can be used to get directly or indirectly an access token.
 //Checks the multi-resource refresh tokens too.
-- (ADTokenCacheItem*)findCacheItemWithKey:(ADTokenCacheKey*) key
-                                        userId:(ADUserIdentifier*)userId
-                                useAccessToken:(BOOL*) useAccessToken
```
jasoncoolmax: I was thinking the same, "Remove it!" :)

jasoncoolmax:
I signed off :)

PR #489, opened 2016-02-04
Synchronization Fix

RPangrle:
This lock was being held too long which could hold up threads in the background.


jasoncoolmax:
I signed off :)

Issue #479, opened 2016-01-29
32-bit Mac Support (No ARC)


RPangrle:
The 32-bit Mac OS runtime does not support ARC, along with some of the more advanced Objective-C literals, however Office for Mac is still 32-bit so we need to support it.

PR #491, opened 2016-02-09
32-bit Mac Support

RPangrle:
Ugh. Legacy ObjC runtime.

On top of adding manual memory management (that's all those SAFE_ARC_* macros) I also had to move all ivar declarations to headers.

@synthesize is required, as properties don't synthesize themselves on the legacy runtime.

I removed some abuse of properties as well, as they caused some very peculiar behavior, particularly around memory management.


RPangrle:
@kpanwar You might want to peek at this.

kpanwar:
@RPangrle - I assume you are looking at the travis issues and this is not simply a build failure?

RPangrle:
Already committed a fix, I tried pushing it yesterday but the wifi on the bus wasn't so great.

ADAL/src/ADAuthenticationContext.m
```diff
@@ -118,10 +131,12 @@ - (id)initWithAuthority:(NSString *)authority
     ADTokenCache* cache = [ADTokenCache new];
     [cache setDelegate:delegate];
     
-    return [self initWithAuthority:authority
+    self = [self initWithAuthority:authority
                  validateAuthority:validateAuthority
                         tokenCache:cache
                              error:error];
+    SAFE_ARC_RELEASE(cache);
+    return self;
```
jasoncoolmax: Why we don't need to SAFE_ARC_AUTORELEASE(self) before return?

ADAL/src/ADClientMetrics.m
```diff
@@ -50,31 +50,38 @@ + (ADClientMetrics*) getInstance {
...
         _endpoint = endPoint;
+        SAFE_ARC_RETAIN(_endpoint);
+        SAFE_ARC_RELEASE(_responseTime);
         _responseTime = @"";
+        SAFE_ARC_RETAIN(_responseTime);
+        SAFE_ARC_RELEASE(_correlationId);
         _correlationId = correlationId;
+        SAFE_ARC_RETAIN(_correlationId);
+        SAFE_ARC_RELEASE(_startTime);
         _startTime = [NSDate new];
```
jasoncoolmax: Why we don't need to retain _startTime?

ADAL/src/ADInstanceDiscovery.m
```diff
@@ -149,9 +136,11 @@ -(void) validateAuthority: (NSString*) authority
     NSString* message = [NSString stringWithFormat:@"Attempting to validate the authority: %@; CorrelationId: %@", authority, [correlationId UUIDString]];
     AD_LOG_VERBOSE(@"Instance discovery", correlationId, message);
     
-    ADAuthenticationError* error;
+    authority = [authority lowercaseString];
```
jasoncoolmax: Is there a memory leak here? :P

ADAL/src/cache/ADTokenCacheItem.m
```diff
@@ -48,55 +47,71 @@ - (void)calculateHash
...
+    item->_authority = [_authority copyWithZone:zone];
+    item->_clientId = [_clientId copyWithZone:zone];
+	item->_familyId = [_familyId copyWithZone:zone];
+    item->_accessToken = [_accessToken copyWithZone:zone];
+    item->_accessTokenType = [_accessTokenType copyWithZone:zone];
+    item->_refreshToken = [_refreshToken copyWithZone:zone];
+    item->_expiresOn = [_expiresOn copyWithZone:zone];
+    item->_userInformation = [_userInformation copyWithZone:zone];
+    item->_sessionKey = [_sessionKey copyWithZone:zone];
+	item->_tombstone = [_tombstone mutableCopyWithZone:zone];
```
jasoncoolmax: if _tombstone is nil, we want the copy value also be nil, so should it be 
item->_tombstone = _tombstone ? [_tombstone mutableCopyWithZone:zone] : nil;
?

ADAL/src/urlprotocol/ADCustomHeaderHandler.m
```diff
@@ -20,42 +20,47 @@
...
-        customHeaders = @{
-                         @"x-ms-PkeyAuth":@"1.0",
-                         };
-        customHeadersForSingleUse = [NSMutableDictionary new];
+    if (self == [ADCustomHeaderHandler class])
+    {
+        s_customHeaders =
+            [NSMutableDictionary dictionaryWithDictionary: @{@"x-ms-PkeyAuth":@"1.0"}];
+        SAFE_ARC_RETAIN(s_customHeaders);
+        s_customHeadersForSingleUse = [NSMutableDictionary new];
```
jasoncoolmax: do we need dealloc function in this class to release objects like s_customHeadersForSingleUse?

ADAL/src/ADAuthenticationContext.m
RPangrle: Once we pass it into the authentication context it takes ownership of it and we don't need it anymore so we release it.

ADAL/src/ADInstanceDiscovery.m
RPangrle: lowercaseString doesn't start with copy, alloc, new or mutableCopy so we don't own the reference to it. Similarly we don't own objects passed to us in parameters

ADAL/src/urlprotocol/ADCustomHeaderHandler.m
RPangrle: File statics live for the life time of the app. There's no particular point releasing "on quit" when the OS is just going to reclaim all the memory anyways, you'll only slow down app quit.

ADAL/src/ADAuthenticationContext.m
RPangrle: Sorry, wrong answer. In this case we didn't call alloc (someone else did) so we don't have a retain count to release or autorelease.

jasoncoolmax: Got it :)

ADAL/src/ADInstanceDiscovery.m
jasoncoolmax: Got it! :)

ADAL/src/urlprotocol/ADCustomHeaderHandler.m
jasoncoolmax: I see, make sense :)

jasoncoolmax:
I signed off :)

ADAL/src/ADAuthenticationSettings.m
```diff
@@ -15,11 +15,17 @@
...
 #import "ADAuthenticationSettings.h"
-#import "ADKeychainTokenCache.h"
 
 @implementation ADAuthenticationSettings
 
+@synthesize requestTimeOut = _requestTimeOut;
+@synthesize expirationBuffer = _expirationBuffer;
+
+#if !TARGET_OS_IPHONE
+@synthesize defaultCacheDelegate = _defaultCacheDelegate;
```
kpanwar: Consider renaming to include 'storage' to clear the purpose.

ADAL/src/ADInstanceDiscovery.m
```diff
@@ -174,43 +163,97 @@ -(void) validateAuthority: (NSString*) authority
...
     AD_LOG_VERBOSE(@"Authority Validation Cache", nil, message);
+    return YES;
+}
+
+- (ADAuthenticationError *)processWebReponse:(ADWebResponse *)webResponse
+                               authorityHost:(NSString *)authorityHost
+                               correlationId:(NSUUID *)correlationId
+{
+    NSInteger code = webResponse.statusCode;
+    if (!(code == 200 || code == 400 || code == 401))
```
kpanwar: We should  check if error 400 returns a JSON. If it does not then deserialization would fail

RPangrle: Um, we are... This if statement is if the code is not (200 or 400 or 401)...

ADAL/src/ADAuthenticationSettings.m
RPangrle: :+1:

ADAL/src/public/ADTokenCacheItem.h
```diff
@@ -25,6 +25,20 @@
     Objects of this class are used in the key-based token cache store.
     See the key extraction function for details on how the keys are constructed. */
 @interface ADTokenCacheItem : NSObject<NSCopying , NSSecureCoding>
+{
+    NSUInteger _hash;
+    NSString* _resource;
+    NSString* _authority;
+    NSString* _clientId;
+    NSString* _familyId;
```
kpanwar: where is familyId being set? also, we are using "foci" as a value in the server response.

kpanwar:
Sign off with comments and an issue filed. https://github.com/AzureAD/azure-activedirectory-library-for-objc/issues/498

ADAL/src/public/ADTokenCacheItem.h
RPangrle: The family ID key has changed several times now. That's a separate issue to be handled in another PR.

RPangrle:
#479

PR #493, opened 2016-02-09
Store idtokens on a per-item basis in ADTokenCache.

RPangrle:
This will be necessary for proper multi tenant support.


jasoncoolmax:
I signed off :)

PR #497, opened 2016-02-11
Set endpoint ivar in ADClientMetrics if endpoint is ADFS - crash fix

mfoland14:
In production environments where ADFS is used for on-prem and exchange servers are online the first login after install would cause the application to crash.

This crash was caused by the fact that the endpoint was not set properly on the first authentication in beginClientMetricsRecordEndpoint:::, this would cause endClientMetricsRecord: to set the isPending flag and the next request would call beginClientMetricsRecordForEndpoint and crash when building the dictionary.

Added one test to cover this scenario.


RPangrle:
Can you please resubmit this PR against dev? Thanks.

Issue #500, opened 2016-02-11
Add correlation ID to caching methods


RPangrle:
We don't need to cache the correlation ID, but we do need it for logging purposes so devs can properly see what requests the cache lookups correspond to, especially in heavily multithreaded usage.

Issue #499, opened 2016-02-11
Make sure ADAuthenticationError logs correlation IDs


RPangrle:
Some of our most important log statements happen at ADAuthenticationError creation, however in our correlationID fixing work we didn't push correlation IDs down to enough places to make sure that we are logging correlation IDs when errors are created.

PR #503, opened 2016-02-12
CorrelationID Error Logging

RPangrle:
#500 #499

We needed to push our correlation IDs farther down the stack to ensure that they got logged in error and caching calls.


jasoncoolmax:
I signed off :)

Issue #508, opened 2016-02-15
Crash: ADTokenCacheItem is passing an NSString to ADAuthenticaitonResult as the Correlation ID


RPangrle:
In this bit of code:

```
    if (![NSString adIsStringNilOrBlank:value])
    {
        BOOL isMrrt = [self fillItemWithResponse:response];
        return [ADAuthenticationResult resultFromTokenCacheItem:self
                                           multiResourceRefreshToken:isMrrt
                                                       correlationId:[response objectForKey:OAUTH2_CORRELATION_ID_RESPONSE]];
```

The value coming out of the dictionary is an NSString, although because it's just an untyped 'id' there the compiler lets it through. This can cause a crash if we try to use the correlation ID as a NSUUID (which is what ADAuthenticationResult thinks it's supposed to be.) Currently the only reason we don't crash (that I can tell) is this value isn't actually used after that point, unless you try to call -description on the result.

PR #509, opened 2016-02-15
Fix a case where we were passing in a NSString where we needed a NSUUID

RPangrle:
Add a check on an acquireToken test, and fix a test that erroneously was checking EqualObjects with a string.


jasoncoolmax:
I signed off :)

RPangrle:
#508

PR #510, opened 2016-02-15
Issues/broker dev/prevent nonhttps redirect in webview

### LINKED ISSUE #487: Verify that auth flows cannot be redirected to a non https endpoint
Except for end url

jasoncoolmax:
to fix #487 
1. Checking is added to prevent insecure http redirect in webview;
2. Unit tests are added, which include adding a class *ADTestAuthenticationViewController* in test to mock ADAuthenticationViewController.


ADAL/src/ui/ADWebAuthController.m
```diff
@@ -290,6 +290,17 @@ - (BOOL)webAuthShouldStartLoadRequest:(NSURLRequest *)request
         return NO;
     }
     
+    // redirecting to non-https url is not allowed
+    if ([requestURL hasPrefix: @"http"] && ![requestURL hasPrefix: @"https"])
+    {
```
jasoncoolmax: Should we just put 
if (![requestURL hasPrefix: @"https"])
and move the checking to the end of this function?

There are some subtle difference :P

RPangrle: just "if ([requestURL hasPrefix: @"http"])" would be sufficient. I could see wanting to specifically block anything that's not either one of the allowed types, redirect URI scheme or HTTPS.

ADAL/tests/ADAuthenticationViewControllerTests.m
```diff
@@ -0,0 +1,102 @@
...
+    
+    [self waitSemaphoreWithoutBlockingMainQueue:_dsem];
+    
+}
+
+- (void)waitSemaphoreWithoutBlockingMainQueue:(dispatch_semaphore_t)sem
+{
+    while (dispatch_semaphore_wait(sem, DISPATCH_TIME_NOW))
+    {
+        [[NSRunLoop mainRunLoop] runMode:NSDefaultRunLoopMode beforeDate:[NSDate dateWithTimeIntervalSinceNow:0.1]];
```
RPangrle: You should be able to use [NSDate distantFuture] here,

RPangrle:
:shipit:

PR #512, opened 2016-02-16
Cleaned up links in ADWebAuthController and ADURLProtocol

RPangrle:
Found these using the Leaks instrument on the test app.


jasoncoolmax:
I signed off :)

PR #514, opened 2016-02-16
Disable tests that fail with no network connection

RPangrle:
Randomly failing tests suck. It's better to not run them at all then be comfortable with a test failure not really being a test failure.


jasoncoolmax:
I signed off :)

Issue #364, opened 2015-08-25
Crash in ADInstanceDiscovery.m - [__NSCFString UUIDString]: unrecognized selector sent to instance


szalap:
The crash appears in [ADInstanceDiscovery validateAuthority:correlationId:completionBlock:] (ADInstanceDiscovery.m:149), where it tries to use UUIDString on correlationId object which is somehow not UUID at this point.
It happened only once, so it is not very devastating, but still.

RPangrle:
This appears to be a synchronization issue on the correlation ID parameter and should be resolved once we fix #387

RPangrle:
Fixed when we switched the correlationID to be a per-request property.

PR #517, opened 2016-02-16
Protect correlationId with synchronization directive

RPangrle:
Mitigation for #364. We have a better fix in 2.1, but those changes are far too wide ranging to bring down to 1.2.x


jasoncoolmax:
I signed off :)

PR #526, opened 2016-02-18
ADURLProtocol correlation IDs

RPangrle:


kpanwar:
:shipit:

kpanwar:
there are still some logs in adal that leak out PII in verbose. this is not your change, but just an FYI

RPangrle:
PII is okay as long as it's in additional info. If we logged no PII at all it'd be extremely difficult to debug some issues.

Issue #523, opened 2016-02-16
Age out the tombstones


jasoncoolmax:
Delete tombstones if they are more than 30 days old.

PR #527, opened 2016-02-18
Issues/dev/age out tombstones

jasoncoolmax:
addressing #523 

1. Changes are made to delete tombstones which are more than 30-day old.
Deletion of out-of-date tombstones is done on the back of the filtering function. Any method involving filtering tombstones will have the effect of deleting old tombstones. 
E.g., [ADKeychainTokenCache allitems: error:] will filter out tombstones and therefore old tombstones are deleted.

2. Unit test for tombstone aging out is added.

3. Fixed the issue that method [ADAuthenticationContext findFamilyItemForUser:] will return tombstone items.

### DIFF
--- a/ADAL/src/cache/ADTokenCache.m
+++ b/ADAL/src/cache/ADTokenCache.m
@@ -205,7 +205,10 @@
                key:(nonnull ADTokenCacheKey *)key
 {
     ADTokenCacheItem* item = [dictionary objectForKey:key];
-    if (item)
+    //if an item is a tombstone and too old, it is deleted and won't be returned;
+    //otherwise it will be returned.
+    if (item &&
+        !([item tombstone] && [self deleteTombstoneIfTooOld:item]))
     {
         item = [item copy];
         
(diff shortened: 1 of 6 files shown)


ADAL/src/cache/ADTokenCacheItem+Internal.m
```diff
@@ -221,6 +221,9 @@ - (void)makeTombstone:(NSDictionary *)tombstoneEntries
     
     //wipe out the refresh token
     _refreshToken = @"<tombstone>";
+    SAFE_ARC_RELEASE(_expiresOn);
+    _expiresOn = [NSDate dateWithTimeIntervalSinceNow:60*60*24*30];//tombstones should be removed after 30 days
```
RPangrle: Put this number in a #define somewhere. #define THIRTY_DAYS_IN_SECONDS

ADAL/src/cache/ADTokenCache.m
```diff
@@ -356,11 +356,29 @@ -(NSMutableArray*)filterOutTombstones:(NSArray*) items
         {
             [itemsKept addObject:item];
         }
+        else
```
RPangrle: Add an age-out check in getItemsWithKey:UserId: as well

jasoncoolmax: I realised that getItemsWithKey:userId: is a better place to add the tombstone deletion code! :)

PR #530, opened 2016-02-19
Add error domains for some specific sources of errors.

RPangrle:
(HTTP codes, Keychain, OAuth Server Errors)

This should make it a little easier to tell what is causing the error and pinpoint more interesting errors.
Remove a duplicative RT log statement
Add a acquire token succeeded log statement.


RPangrle:
@kpanwar

ADALiOS/ADALiOS/ADKeyChainHelper.m
```diff
@@ -133,21 +126,9 @@ -(BOOL) updateItemByAttributes: (NSDictionary*) attributes
             //All good
             return YES;
         case errSecItemNotFound:
-        {
```
jasoncoolmax: In this case, may I know the reason that we return YES for the update call? :)

jasoncoolmax:
I signed off :)

PR #533, opened 2016-02-20
Move TokenCaching category into AuthenticationRequest

RPangrle:
When I created ADAuthenticationRequest I did it almost entirely off of categories that I had separated out from ADAuthenticationContext and then switched the classes of. I had left TokenCaching alone at the time because it appeared to dip into properties of the authentication context too much. Now with some of the logging changes I want to make, and some of the changes we've already made regarding correlation IDs, it makes sense to move it into ADAuthenticationRequest as well.


jasoncoolmax:
I signed off :)

Issue #490, opened 2016-02-09
Check if we're in an extension before popping UI


RPangrle:
We can't pop UI or go to broker in an extension, so if we're going to go to an interactive flow we should detect this case and give a more coherent error.

```
if ([[[NSBundle mainBundle] bundlePath] hasSuffix:@".appex"]) {
    // this is an app extension
}
```

RPangrle:
We could add this after the silent check in requestToken:

PR #535, opened 2016-02-23
App extension check is added

jasoncoolmax:
(Addressing #490)
Checking is added to make sure we will not launch webview/broker if it is in an app extension.


RPangrle:
:shipit:

Issue #541, opened 2016-02-29
Developers' own UIWebView fails to return after completion


jasoncoolmax:
When developers sent their own UIWebView, after on completion of that login, it fails to return.
(Version 2.1.0 Beta 3)

Issue #539, opened 2016-02-29
App crashes when developers set their own UIWebView


jasoncoolmax:
When an app gives its own UIWebView to be used for ADAL login (by setting ADAuthenticationContext.webView), on completion of that login a message is getting sent to a deallocated ADAuthenticationViewController and crashes the app.
(Version 2.1.0 Beta 3)

Issue #538, opened 2016-02-29
Passing in NSString* instead of NSUUID* will crash ADAL


jasoncoolmax:
Notice ADAL crash when NSString* is passed in as CorrelationId where NSUUID* is expected.
It happens in ADAuthenticationContext+Internal.m.

Need to throughly check other files to make sure it won't happen in other files.

PR #542, opened 2016-02-29
Bug fixed for ADAL crash when developers set their own UIWebView

jasoncoolmax:
Bug fixes for #539, #541 and #538 

1) ADAL crashes when developers set their own UIWebView because the webview's delegate property hasn't been nil out when webview controller is deallocated.

2) Fix has been made for the bug that developers' own UIWebView fails to return after completion.

3) Fixed a place where NSString* is passed in as CorrelationId but NSUUID* is expected.

### DIFF
--- a/ADAL/src/ui/ios/ADAuthenticationViewController.m
+++ b/ADAL/src/ui/ios/ADAuthenticationViewController.m
@@ -108,6 +108,7 @@ NSString *const AD_FAILED_NO_CONTROLLER = @"The Application does not have a curr
 -(void)dealloc
 {
     [_webView setDelegate:nil];
+    _webView = nil;
 }
 
 #pragma mark - UIViewController Methods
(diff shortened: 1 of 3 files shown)


ADAL/src/ui/ios/ADAuthenticationViewController.m
```diff
@@ -103,6 +103,13 @@ - (BOOL)loadView:(ADAuthenticationError * __autoreleasing *)error
     return NO;
 }
 
+/*! set webview's delegate to nil when the view controller 
+    is deallocated, or it might crash ADAL. */
+-(void)dealloc
+{
```
RPangrle: Also nil out _webView here as well

jasoncoolmax: Sure :)

RPangrle:
Also take a look at the mac side for any similar issue.

jasoncoolmax:
I cannot reproduce the crash for either 32-bit or 64-bit mac.
Do we want to add the dealloc to mac anyway?

RPangrle:
:shipit:

Issue #543, opened 2016-03-02
Client ID is not replaced for scope value in request


jasoncoolmax:
Client ID should be removed, "openid" and "offline_access" should be added for scope value before sending the request.

PR #546, opened 2016-03-02
Remove Client ID from scope in request.

jasoncoolmax:
(Fix for #543)

Previously, If client ID is in the scope, we keep it in scope and won't add "openid"/"offline_access". 

Now we change it:
Client ID will always removed from scope (if any); "openid" and "offline_access" will be added to scope.


jasoncoolmax:
@brentschmaltz , could you please also look at this? :)

RPangrle:
Can you target this to covergence-dev instead?

I'm assuming this is due to some spec change. This behavior was copied from the .NET code (I think) 
@kpanwar might want to chime in.

jasoncoolmax:
Sure :)

jasoncoolmax:
Changes have been targetted at convergence-dev branch. So close this PR.

PR #547, opened 2016-03-05
Removed Client ID from scope in request.

jasoncoolmax:
(Fix for #543)

Previously, If client ID is in the scope, we keep it in scope and won't add "openid"/"offline_access".

Now we change it:
Client ID will always removed from scope (if any); "openid" and "offline_access" will be added to scope.


RPangrle:
@kpanwar Can you please chime in and let us know if this is something that was fixed in other ADALs.

kpanwar:
Change looks good. I made sure that you are doing necessary checks in validateScopes() as well. :shipit:

PR #559, opened 2016-03-14
Copy over some logging changes from 2.0.3

RPangrle:


ADAL/src/ADAuthenticationError.m
```diff
@@ -243,5 +246,45 @@ + (ADAuthenticationError*)errorFromNonHttpsRedirect:(NSUUID *)correlationId
...
+                                description:(NSString *)description
+                                       code:(NSInteger)code
+                              correlationId:(NSUUID *)correlationId
+{
+    return [self errorWithDomainInternal:ADOAuthServerErrorDomain
+                                    code:code
+                       protocolErrorCode:protocolCode
+                            errorDetails:description
+                           correlationId:correlationId
+                                userInfo:nil];
```
kpanwar: Are we not populating userInfo with details yet?

RPangrle: This is an authentication error, not a log statement, and all of those details are accessible as properties already. userInfo is typically for extra information, which we'll eventually use for chaining errors

kpanwar:
:shipit:

PR #568, opened 2016-03-24
Categorize and renumber error codes

RPangrle:


ADAL/src/public/ADErrorCodes.h
```diff
@@ -28,101 +28,133 @@ typedef enum
      but no error condition occurred.*/
     AD_ERROR_SUCCEEDED = 0,
     
-    /*! The user has cancelled the applicable UI prompts */
-    AD_ERROR_USER_CANCEL = 1,
+    /*! An unexpected internal error occurred. */
+    AD_ERROR_UNEXPECTED = -1,
```
jasoncoolmax: Seems it is made negative on purpose, just to get your attention in case it is not :P

jasoncoolmax:
I signed off :)

RandalliLama:
:shipit:

PR #570, opened 2016-03-25
Main thread checks.

RPangrle:
Check in -acquireToken: to make sure we're on the main thread. In ADAL 2.1 we require interactive auth to be initiated on the main thread as the developer is requesting a UI operation. However it is still possible that we fail a RT attempt and end up on a network thread before popping UI so we still need to jump back to the main thread in that case.


kpanwar:
:SHIPIT:

jasoncoolmax:
I signed off :)

PR #574, opened 2016-03-28
Keychain Back compat + Version Info

RPangrle:
* Changes to match the keychain lookup logic on 1.2.x
* Add class name mapping to ADTokenCacheItem so that the items serialize and deserialize the same.
* Add some redundant information in ADUserInformation because 1.2.x doesn't bother reparsing the idtoken.
* userId was base64 encoded in the account field in 1.2.x
* Add an NSLog at startup to print out the ADAL version used, from a string generated off of the version.
* Add a symbol in the ADAL binary that matches the ADAL version, so we can inspect a binary without having to run it to discover the version.


jasoncoolmax:
I signed off :)

Issue #550, opened 2016-03-10
Grab the interactive lock before calling broker


RPangrle:
If for some reason multiple interactive auth attempts are happening at the same time they both could end up trying to call broker and weird behavior will result.

PR #577, opened 2016-03-30
Exclusion lock is now grabbed before calling broker

jasoncoolmax:
(for #550)
Before calling broker, an exclusion lock is grabbed.

The exclusion lock used for calling broker is the same as the one used for popping webview, as we want to make sure one interactive request is going on at a time.

### DIFF
--- a/ADAL/src/request/ADAuthenticationRequest+WebRequest.m
+++ b/ADAL/src/request/ADAuthenticationRequest+WebRequest.m
@@ -356,11 +356,7 @@ static ADAuthenticationRequest* s_modalRequest = nil;
 //Attempts to release the lock. Logs warning if the lock was already released.
 -(void) releaseExclusionLock
 {
-    if ( ![self releaseUserInterationLock] )
-    {
-        AD_LOG_WARN(@"UI Locking", _correlationId, @"The UI lock has already been released.");
-    }
-    
+    [self releaseUserInterationLock];
     s_modalRequest = nil;
 }
 
--- a/ADAL/src/request/ADAuthenticationRequest+WebRequest.m
+++ b/ADAL/src/request/ADAuthenticationRequest+WebRequest.m
@@ -332,16 +332,13 @@ static ADAuthenticationRequest* s_modalRequest = nil;
     SAFE_ARC_RELEASE(webRequest);
 }
 
-//Used for the callback of obtaining the OAuth2 code:
-static volatile int sDialogInProgress = 0;
-
 //Ensures that a single UI login dialog can be requested at a time.
 //Returns true if successfully acquired the lock. If not, calls the callback with
 //the error and returns false.
 - (BOOL)takeExclusionLockWithCallback: (ADAuthorizationCodeCallback) completionBlock
 {
     THROW_ON_NIL_ARGUMENT(completionBlock);
-    if ( !OSAtomicCompareAndSwapInt( 0, 1, &sDialogInProgress) )
+    if ( ![self takeUserInterationLock] )
     {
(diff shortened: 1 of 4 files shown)


ADAL/src/request/ADAuthenticationRequest+WebRequest.m
```diff
@@ -332,16 +332,13 @@ - (void)requestWithServer:(NSString *)authorizationServer
     SAFE_ARC_RELEASE(webRequest);
 }
 
-//Used for the callback of obtaining the OAuth2 code:
-static volatile int sDialogInProgress = 0;
-
 //Ensures that a single UI login dialog can be requested at a time.
 //Returns true if successfully acquired the lock. If not, calls the callback with
 //the error and returns false.
 - (BOOL)takeExclusionLockWithCallback: (ADAuthorizationCodeCallback) completionBlock
```
jasoncoolmax: I found that this function cannot be directly used for the broker case, because it accepts a different completionBlock type and do additional things in the function :P

ADAL/src/request/ADAuthenticationRequest.m
```diff
@@ -268,4 +268,14 @@ - (NSUUID*)correlationId
     return _correlationId;
 }
 
+- (BOOL)takeUserInterationLock
+{
+    return OSAtomicCompareAndSwapInt( 0, 1, &sInteractionInProgress);
```
RPangrle: Use a NSLock or a Dispatch primitive of some sort.

jasoncoolmax: I tried to use NSLock, but I found an interesting thing: the reference says 
"When sending an unlock message to an NSLock object, you must be sure that message is sent from the same thread that sent the initial lock message. Unlocking a lock from a different thread can result in undefined behavior."

We may need to release a lock in a completion block, where the thread executing the block is uncertain :P 

For "dispatch primitive", do you mean dispatch_once? :)

RPangrle: It's possible to use a dispatch_semaphore as a lock. Just have it return immediately on the wait and check whether or not you got the semaphore to try and get the lock and signal to let go of the lock.

jasoncoolmax: Sure :)

ADAL/src/request/ADAuthenticationRequest.m
```diff
@@ -36,6 +36,10 @@
 
 #include <libkern/OSAtomic.h>
 
+// Used to make sure one interactive request is going on at a time,
+// either launching webview or broker
+static volatile dispatch_semaphore_t sInteractionInProgress = nil;
```
RPangrle: remove volatile keyword.

jasoncoolmax: Sure

ADAL/src/request/ADAuthenticationRequest.m
```diff
@@ -85,6 +89,16 @@ + (ADAuthenticationRequest*)requestWithContext:(ADAuthenticationContext*)context
     return request;
 }
 
+- (id)init
+{
+    if (!(self = [super init]))
+        return nil;
+    
+    sInteractionInProgress = dispatch_semaphore_create(1);
```
RPangrle: This is not the right place to initialize a static variable. You'll end up recreating the semaphore every time a new ADAuthenticationRequest is created.

Put this in + (void)initialize

jasoncoolmax: Oh, my bad :P

RPangrle:
:shipit:

PR #579, opened 2016-03-31
Fix method type handling

RPangrle:
The use of global variables was not in line with best practices, and the overall handling of that ivar was incorrect in some cases, but we were most likely getting away with it due to it only being set with constant compile time defined NSStrings. I'm cleaning up the interface a bit here to make it less likely this could cause issues.


RPangrle:
#557

jasoncoolmax:
:shipit:

Issue #582, opened 2016-04-01
Policy value is not set before storing token items to cache


jasoncoolmax:
When an token item is stored to keychain cache store, the policy value is not set. The consequence of it is, ADAL will resort to webview every time because it cannot retrieve AT/RT properly from cache.

PR #584, opened 2016-04-05
Policy is now set before a cache item is stored.

jasoncoolmax:
(For #582)

Policy value is used to look up for proper refresh token from cache but previously it was not set for each cache item.
Now policy value is set before a cache item is stored.

--------------

When adding the unit test, I found a few more bugs:
1) For [ADKeychainTokenCacheStore getItemWithKey:error:], policy value is not retrieved.

2) userId being NSNull should be treated the same as being nil. Otherwise it can cause ADAL to crash in one place, it can also cause cache to unwanted behaviour.

### DIFF
--- a/ADALiOS/ADALiOS/ADKeychainTokenCacheStore.m
+++ b/ADALiOS/ADALiOS/ADKeychainTokenCacheStore.m
@@ -259,7 +259,11 @@ static void adkeychain_dispatch_if_needed(dispatch_block_t block)
     }
     
     ADKeychainQuery* writeQuery = [self createBaseQuery];
-    [writeQuery setUserId:userId ? userId : @""];
+    if (!userId || [userId isKindOfClass:[NSNull class]])
+    {
+        userId = @"";
+    }
+    [writeQuery setUserId:userId];
     
     const void * keys[] = { kSecAttrGeneric };
     const void * values[] = { data };
--- a/ADALiOS/ADALiOS/ADUserIdentifier.m
+++ b/ADALiOS/ADALiOS/ADUserIdentifier.m
@@ -97,7 +97,7 @@
     }
     
     NSString* matchString = [identifier userIdMatchString:info];
-    if (!matchString || [matchString isEqualToString:identifier.userId])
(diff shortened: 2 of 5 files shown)


RPangrle:
Are there any unit tests that can be added that would have caught this?

jasoncoolmax:
Let me check ;)

ADALiOS/ADALiOS/ADUserIdentifier.m
```diff
@@ -97,7 +97,7 @@ + (BOOL)identifier:(ADUserIdentifier*)identifier
     }
     
     NSString* matchString = [identifier userIdMatchString:info];
-    if (!matchString || [matchString isEqualToString:identifier.userId])
```
jasoncoolmax: matchString can be NSNull here. This line crashes in some cases.

ADALiOS/ADALiOSTests/ADAuthenticationContextTests.m
```diff
@@ -415,6 +415,7 @@ - (void)asynchronousAcquireTokenWithLine:(int)line
                                   identifier:[ADUserIdentifier identifierWithId:_userId type:RequiredDisplayableId]
                                promptBehavior:_promptBehavior
                            extraQueryParameters:nil
+                                       policy:_policy
```
jasoncoolmax: previously acquireTokenWithScopes with policy parameter is not test.

ADALiOS/ADALiOS/ADKeychainTokenCacheStore.m
```diff
@@ -259,7 +259,11 @@ - (OSStatus)writeDictionary:(CFDictionaryRef)dictionary
     }
     
     ADKeychainQuery* writeQuery = [self createBaseQuery];
-    [writeQuery setUserId:userId ? userId : @""];
```
jasoncoolmax: userId could be NSNull here. We need to replace nil/NSNull with @""

RPangrle:
:shipit:

PR #589, opened 2016-04-08
iOS Test App Overhaul

RPangrle:
Gone are the iPhone vs iPad specific storyboards, and storyboards all together. Some UI is layed out in XIBs, some is built programatically depending on the complexity of the UI.
You can now view and fiddle with the cache, including delete, tombstone and expire items as appropriate!
This test app allows testing both the provided ADAL UI and passed in webviews.


jasoncoolmax:
:shipit:

PR #592, opened 2016-04-08
Error double pointers must be checked before use.

RPangrle:


ADAL/src/cache/ios/ADKeychainTokenCache.m
```diff
@@ -298,14 +298,22 @@ - (BOOL)removeAllForClientId:(NSString * __nonnull)clientId
...
         if ([clientId isEqualToString:[item clientId] ])
         {
-            [self removeItem:item error:error];
-            if (*error)
+            [self removeItem:item error:&adError];
+            if (adError)
             {
                 deleteSuccessful = NO;
+                
+                if (error)
```
jasoncoolmax: Question: why doesn't the __nullable keyword guarantee *if (error)* being true here? :P

jasoncoolmax:
:shipit:

PR #593, opened 2016-04-09
Changes made to age out tombstone.

jasoncoolmax:
(for #523)

Tombstones 30-day old will be deleted.
Clean frequency is set to be once a day. Cleaning necessity check is performed once per app launch, i.e. the checking code is put at the init function of ADKeychainTokenCache using dispatch_once.

### DIFF
--- a/ADAL/src/cache/ADTokenCacheItem+Internal.m
+++ b/ADAL/src/cache/ADTokenCacheItem+Internal.m
@@ -34,7 +34,7 @@
 @implementation ADTokenCacheItem (Internal)
 
 #define CHECK_ERROR(_CHECK, _ERR) { if (_CHECK) { if (error) {*error = _ERR;} return; } }
-#define THIRTY_DAYS_IN_SECONDS 2592000
+#define THIRTY_DAYS_IN_SECONDS (30*24*60*60)
 
 - (void)checkCorrelationId:(NSDictionary*)response
       requestCorrelationId:(NSUUID*)requestCorrelationId
--- a/ADAL/src/cache/ADTokenCacheItem+Internal.m
+++ b/ADAL/src/cache/ADTokenCacheItem+Internal.m
@@ -34,6 +34,7 @@
 @implementation ADTokenCacheItem (Internal)
 
 #define CHECK_ERROR(_CHECK, _ERR) { if (_CHECK) { if (error) {*error = _ERR;} return; } }
+#define THIRTY_DAYS_IN_SECONDS 2592000
 
 - (void)checkCorrelationId:(NSDictionary*)response
       requestCorrelationId:(NSUUID*)requestCorrelationId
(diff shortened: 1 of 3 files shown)


ADAL/src/cache/ADTokenCacheItem+Internal.m
```diff
@@ -34,6 +34,7 @@
 @implementation ADTokenCacheItem (Internal)
 
 #define CHECK_ERROR(_CHECK, _ERR) { if (_CHECK) { if (error) {*error = _ERR;} return; } }
+#define THIRTY_DAYS_IN_SECONDS 2592000
```
RPangrle: Make this "30 * 24 * 60 * 60" so it's easier to visually inspect. The compiler will multiply all of that at build time for you.

ADAL/src/cache/ios/ADKeychainTokenCache.m
```diff
@@ -35,12 +35,19 @@
 #define KEYCHAIN_VERSION 1
 #define STRINGIFY(x) #x
 #define TOSTRING(x) STRINGIFY(x)
+#define ONE_DAY_IN_SECONDS 86400
```
RPangrle: Similarly, change to "24 * 60 * 60"

ADAL/src/cache/ios/ADKeychainTokenCache.m
```diff
@@ -35,12 +35,19 @@
...
 
 static NSString* const s_nilKey = @"CC3513A0-0E69-4B4D-97FC-DFB6C91EE132";//A special attribute to write, instead of nil/empty one.
 static NSString* const s_delimiter = @"|";
 
 static NSString* const s_libraryString = @"MSOpenTech.ADAL." TOSTRING(KEYCHAIN_VERSION);
 
+// artificial key made up for storing the next tombstone clean time in keychain
```
RPangrle: Don't store this as an artificial token, that'll trip up old versions and confuse developers that are going through the allItems array. This should be a separate keychain entry independent of the tokens themselves.

ADAL/src/cache/ios/ADKeychainTokenCache.m
```diff
@@ -35,12 +35,15 @@
...
 #define STRINGIFY(x) #x
 #define TOSTRING(x) STRINGIFY(x)
+#define ONE_DAY_IN_SECONDS (24*60*60)
 
 static NSString* const s_nilKey = @"CC3513A0-0E69-4B4D-97FC-DFB6C91EE132";//A special attribute to write, instead of nil/empty one.
 static NSString* const s_delimiter = @"|";
 
 static NSString* const s_libraryString = @"MSOpenTech.ADAL." TOSTRING(KEYCHAIN_VERSION);
 
+static NSString* const s_keyForStoringTomestoneCleanTime = @"NEXT-TOMBSTONE-CLEAN-TIME";
```
RPangrle: Preface this with "ADAL" just in case.

RPangrle:
:shipit:

PR #594, opened 2016-04-12
Improve ADLogger header

RPangrle:
We weren't properly documenting the full purpose of all the strings in the ADLogger callback. This should decrease confusion on that.


jasoncoolmax:
:shipit:

PR #596, opened 2016-04-13
Remove the Framework creation script hack

RPangrle:
This predates actual framework support on iOS. It wasn't well formed and caused the archive step to fail, so it's a safe bet that no one is relying on it anyways. Anyone using ADALiOS as a framework is almost certainly doing it via CocoaPods, which has its own problems but at least mostly "works."


RPangrle:
Sigh. I forgot about the dependencies in the test app on this hack. I guess people will have to wait until 2.1

PR #597, opened 2016-04-13
Invalidate MRRTs and fix Xcode Projects

RPangrle:
* Add an "Invalidate" option to MRRTs that replaces the MRRT with "<bad-refresh-token>" which the server will then reject
* Fix the Xcode projects so that the test app can be deployed to device via Xcode.


jasoncoolmax:
:shipit:

PR #599, opened 2016-04-14
Test App Versioning

RPangrle:
* Set the display name to the ADAL version
* Change the bundle id to "com.microsoft.adal.2.1.0.TestApp (Changing bundle IDs with version lets us have test apps for multiple ADAL versions side-by-side)
* Add an app icon in a shade of purple
* Set the app tint color to match the icon color.
* Set the default redirect URI and client ID to match an application listed in the msdevex tenant
* Fix a UI issue where the UI wasn't updating in a timely manner once we were done signing in.


jasoncoolmax:
:shipit:

PR #601, opened 2016-04-14
Change generic attr on tombstone timestamp

RPangrle:
The tombstone timestamp is tripping up old versions of ADAL and causing allItems to fail on 1.2.x. Change the generic on it so it doesn't show up in their keychain queries.


jasoncoolmax:
:shipit:

PR #602, opened 2016-04-14
Keychain Team ID and WPJ State on Test App Settings screen

RPangrle:
This should make it easier to double check that the app is in a known state. Because I'm dipping into more internal APIs here I bit the bullet and setup the test app to go straight at the internal headers. At some point in the future we should setup a "Sample" app that demos just the public APIs.


jasoncoolmax:
:shipit:

PR #603, opened 2016-04-14
Check Broker RedirectURI in -acquireToken:

RPangrle:
If the authentication context is set up to use broker then we want to fail really early on an invalid redirect URI, before we check for the broker being installed. This way we don't surprise developers if their code doesn't work when the Authenticator is installed.


jasoncoolmax:
:shipit:

PR #620, opened 2016-04-25
Packed underlying errors to the error object when interaction needed.

jasoncoolmax:
For #580

Underlying error is stored in the error object when AD_ERROR_SERVER_USER_INPUT_NEEDED error is returned.

### DIFF
--- a/ADAL/src/request/ADAuthenticationRequest+AcquireToken.h
+++ b/ADAL/src/request/ADAuthenticationRequest+AcquireToken.h
@@ -32,8 +32,7 @@
 
 // Bypasses the cache and attempts to request a token from the server, generally called after
 // attempts to use cached tokens failed
-- (void)requestToken:(ADAuthenticationError*)previousError
-     completionBlock:(ADAuthenticationCallback)completionBlock;
+- (void)requestToken:(ADAuthenticationCallback)completionBlock;
 
 // Generic OAuth2 Authorization Request, obtains a token from an authorization code.
 - (void)requestTokenByCode:(NSString*)code
--- a/ADAL/src/request/ADAuthenticationRequest+AcquireToken.h
+++ b/ADAL/src/request/ADAuthenticationRequest+AcquireToken.h
@@ -32,7 +32,8 @@
 
 // Bypasses the cache and attempts to request a token from the server, generally called after
 // attempts to use cached tokens failed
-- (void)requestToken:(ADAuthenticationCallback)completionBlock;
+- (void)requestToken:(ADAuthenticationError*)previousError
+     completionBlock:(ADAuthenticationCallback)completionBlock;
 
 // Generic OAuth2 Authorization Request, obtains a token from an authorization code.
 - (void)requestTokenByCode:(NSString*)code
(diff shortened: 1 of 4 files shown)


ADAL/src/request/ADAuthenticationRequest+AcquireToken.m
```diff
@@ -231,6 +232,7 @@ - (void)requestToken:(ADAuthenticationCallback)completionBlock
         [ADAuthenticationError errorFromAuthenticationError:AD_ERROR_SERVER_USER_INPUT_NEEDED
                                                protocolCode:nil
                                                errorDetails:ADCredentialsNeeded
+                                                   userInfo:@{NSUnderlyingErrorKey:(previousError?previousError:[NSNull null])}
```
RPangrle: This key should be entirely missing if you don't have an underlying error.

jasoncoolmax: OK :)

ADAL/src/request/ADAuthenticationRequest+AcquireToken.h
```diff
@@ -32,7 +32,8 @@
 
 // Bypasses the cache and attempts to request a token from the server, generally called after
 // attempts to use cached tokens failed
-- (void)requestToken:(ADAuthenticationCallback)completionBlock;
+- (void)requestToken:(ADAuthenticationError*)previousError
```
RPangrle: Try to minimize parameters on methods unless there's no way around it, otherwise the number of parameters has a tendency to explode. You should use ivars instead when you're in ADAuthenticationRequest

ADAL/src/request/ADAuthenticationRequest+AcquireToken.m
```diff
@@ -130,7 +130,7 @@ - (void)validatedAcquireToken:(ADAuthenticationCallback)completionBlock
         }
     }
     
-    [self requestToken:completionBlock];
+    [self requestToken:error completionBlock:completionBlock];
```
RPangrle: error is always going to be nil in this case.

RPangrle: The place where we're really looking for underlying errors is when the server sent us an OAuth error and we end up stomping on it and returning interaction required instead.

jasoncoolmax: Sure! :)

ADAL/src/request/ADAuthenticationRequest+AcquireToken.m
```diff
@@ -227,11 +229,18 @@ - (void)requestToken:(ADAuthenticationCallback)completionBlock
         //The cache lookup and refresh token attempt have been unsuccessful,
         //so credentials are needed to get an access token, but the developer, requested
         //no UI to be shown:
-        ADAuthenticationError* error =
+        ADAuthenticationError* error = _underlyingError ?
```
RPangrle: Move the dictionary creation into a separate variable and just pass that into the ADAuthenticationError initializer

jasoncoolmax: I was thinking how to simplify this part. What you said is what I was looking for! :)

RPangrle:
:shipit:

PR #623, opened 2016-04-26
Only delete RTs when we get 'invalid_grant' from the server

RPangrle:
MRRTs might still be good for other resources, so we want to limit the cases where we delete them. We still are working with the service to reduce the cases where they send us invalid_grant when another error is more appropriate.

#606


jasoncoolmax:
:shipit:

PR #631, opened 2016-05-04
Keychain Cache Helper APIs

RPangrle:
Add convenience methods for grabbing the default keychain cache.
Add an exception when the default keychain group is changed improperly.


kpanwar:
:shipit:

Issue #611, opened 2016-04-20
Skip Install Setting


5fcgdaeb:
Hello,

I am trying to archive my app that uses the azure-activedirectory-library-for-objc framework (master branch, release 2.1). I am not using Cocoapods to integrate with the library; I did it manually.

When I try to archive the project in XCode (it basically means create a distribution build), it gets built as 'Generic XCode Archive' as opposed to an 'iOS App Archive'.

I did some analysis and if I change the 'Skip Install' build setting on the ADAL project to 'YES', the archiving works as intended. Am I missing something with regards to project setup? Would it be possible to change this setting?

jasoncoolmax:
@5fcgdaeb This is an issue of XCode rather than ADAL :)
According to Apple's documents, seems you are doing the right thing setting the "skip install" to 'YES' for static library/framework (https://developer.apple.com/library/ios/technotes/tn2215/_index.html#//apple_ref/doc/uid/DTS40011221-CH1-PROJ)

Please let us know whether your problem has been solved :)

5fcgdaeb:
@jasoncoolmax Thanks for your input. The problem is that this setting is set to 'NO' in ADAL framework itself (not a setting that I can change). I need to change this one to 'YES' in ADAL to make the archiving work. Shouldn't it come with the setting set as 'NO' ?

RPangrle:
SKIP_INSTALL is set to 'YES" for the iOS Framework:
https://github.com/AzureAD/azure-activedirectory-library-for-objc/blob/master/ADAL/ADAL.xcodeproj/project.pbxproj#L1795

5fcgdaeb:
That is correct but ADAL target has a target dependency on ADALiOS target.  Could that be causing the issue?

Here is what I have done:
1) Check out master branch.
2) Archive -> Generic XCode Archive
3) Change the Skip Install setting of ADALiOS target to 'YES'.
4) Archive -> iOS App Archive

![screen shot 2016-04-27 at 13 16 37](https://cloud.githubusercontent.com/assets/877496/14849194/4dab5f0c-0c7a-11e6-805a-92ad7010e526.png)

Am I the only one experiencing this issue? Can I switch to another version/tag where this is working?

jasoncoolmax:
@5fcgdaeb thank you very much for your feedback! Issue has been filed and we will fix that :)
https://github.com/AzureAD/azure-activedirectory-library-for-objc/issues/624

[...]

PR #634, opened 2016-05-06
Set build setting SKIP_INSTALL to YES for static library and framework

jasoncoolmax:
for #611

Build setting SKIP_INSTALL is now set to YES for both static library and framework.
The above setting has been moved from project file into the xcconfig files.


RPangrle:
:shipit:

PR #639, opened 2016-05-11
Make macro CHECK_PROFILE_INFO_ERROR more robust

jasoncoolmax:
XCode 7.3 is giving an error message "Logical not is only applied to the left hand side of this comparison" for the above macro.

Fixed it by adding parenthesis.


RPangrle:
:shipit:

PR #641, opened 2016-05-13
ADTokenCacheItem forward compatible extensibility

RPangrle:
Add dictionaries for additional properties coming from the server or client so that we don't have to worry about old client clobbering new properties anymore if they're sharing the same cache and everyone is on ADAL 2.2 or greater.

Also, any properties we're not currently caching coming from the server we'll store in "additionalServer" so regardless of the version of ADAL involved we will retain any information we get from the server so that it might potentially be used in later versions of ADAL.

### DIFF
--- a/ADAL/src/cache/ADTokenCacheItem.m
+++ b/ADAL/src/cache/ADTokenCacheItem.m
@@ -79,6 +79,8 @@
     item->_userInformation = [_userInformation copyWithZone:zone];
     item->_sessionKey = [_sessionKey copyWithZone:zone];
 	item->_tombstone = [_tombstone mutableCopyWithZone:zone];
+    item->_additionalClient = [_additionalClient mutableCopyWithZone:zone];
+    item->_additionalServer = [_additionalServer copyWithZone:zone];
     
     [item calculateHash];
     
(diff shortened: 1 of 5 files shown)


RPangrle:
@sdabbiru @weijjia @jasoncoolmax @heidijinxujia

ADAL/src/cache/ADTokenCacheItem.m
```diff
@@ -189,6 +193,10 @@ - (id)initWithCoder:(NSCoder *)aDecoder
     SAFE_ARC_RETAIN(_userInformation);
 	_tombstone = [aDecoder decodeObjectOfClass:[NSMutableDictionary class] forKey:@"tombstone"];
 	SAFE_ARC_RETAIN(_tombstone);
```
addev-ashish: not a part of this change, but i see on and off tabs in this file.

RPangrle: Someone there used tabs instead of spaces and github is rendering a tab as 8 spaces instead of 4. :-/ Perhaps at some point in the future we'll do a tabs -> spaces pass, but not in this change.

jasoncoolmax:
I signed off :)

kpanwar:
:shipit:

PR #647, opened 2016-05-25
B2C fixes: remove reserved scope values for cache lookup

jasoncoolmax:
make the following fixes:

1) during cache item lookup, the following four scope values are now removed for lookup purpose. Because server will not return such scopes and they are not stored with the cache item.

*offline_access, openid, profile, email*

2) fix a bug where displayable userId is not lowercased for comparison.

### DIFF
--- a/ADALiOS/ADALiOS/ADAuthenticationRequest.m
+++ b/ADALiOS/ADALiOS/ADAuthenticationRequest.m
@@ -279,7 +279,7 @@ static NSMutableArray* _arrayOfLowercaseStrings(NSArray* strings, NSString* cont
                                          uniqueId:uniqueId
                                            idType:_identifier.type
                                            policy:_policy
-                                           scopes:_scopes
+                                           scopes:[self scopesWithoutReservedValues:_scopes]
                                             error:error];
 }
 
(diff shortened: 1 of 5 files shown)


ADALiOS/ADALiOSTests/ADAuthenticationContextTests.m
```diff
@@ -1054,7 +1054,7 @@ - (void)testExtractAndStoreTokenFromResponse
                                                      dictionaryAsJSON:@{ OAUTH2_TOKEN_TYPE : @"Bearer",
                                                                          OAUTH2_REFRESH_TOKEN : newRefreshToken,
                                                                          OAUTH2_ID_TOKEN : newIdToken,
-                                                                         OAUTH2_SCOPE : @"plantetarydefense.fire planetarydefense.target openid offline_access"
+                                                                         OAUTH2_SCOPE : @"plantetarydefense.fire planetarydefense.target"
                                                                          }];
```
jasoncoolmax: this unit test is changed accordingly.

ADALiOS/ADALiOS/ADAuthenticationRequest.m
```diff
@@ -311,5 +311,14 @@ - (BOOL)validateProperties:(ADAuthenticationCallback)completionBlock
     return YES;
 }
 
+- (NSSet*)scopesWithoutReservedValues:(NSSet*)scopes
```
RPangrle: You're always calling this with _scopes, the parameter is unnecessary

ADALiOS/ADALiOS/ADAuthenticationRequest.m
```diff
@@ -311,5 +311,14 @@ - (BOOL)validateProperties:(ADAuthenticationCallback)completionBlock
...
+{
+    NSMutableSet* newScopes = [scopes mutableCopy];
```
RPangrle: Put this in an ivar in the request instead of creating the set anew each time.

jasoncoolmax: Sure :)

RPangrle:
:shipit:

PR #648, opened 2016-05-25
Copy PKeyAuth fix from 2.x

RPangrle:
Copy over the fix to use either cert issuer or thumbprint, depending on what's specified in the auth challenge.


jasoncoolmax:
:shipit:

kpanwar:
:shipit:

PR #650, opened 2016-05-26
Add resiliency feature

jasoncoolmax:
To achieve uninterrupted services access even during an Evo STS outage, server may return ext_expires_in value to indicate the extended lifetime of an access token.

The following changes are made to support the resiliency feature:

1) In ADAuthenticationContext,
Property **extendedLifetimeEnabled** is added for developers to opt in the feature.

2) In ADAuthenticationResult,
Property **extendedLifeTimeToken** is added to indicate whether the token being returned is a token only valid in terms of extended lifetime.

3) In ADAcquireTokenSilentHandler, _(token return logic)_
Through out the normal cache lookup sequence(getAccessToken, tryRT, tryMRRT, etc.), we store the valid access token in terms of ext_expires_in (if we find any) to a variable **_validStaleAccessTokenItem**. In the end of the lookup sequence, we will return **_validStaleAccessTokenItem** as result if *extendedLifetimeEnabled* is on and server is unavailable.

4) In ADAcquireTokenSilentHandler, _(token delete logic)_
An access token will be deleted only if the token expires in terms both _expires_in_ and _ext_expires_in_.

5) In ADWebAuthRequest,
Now it will retry once after a second if server response is 500/503/504. (previously only retry for 500/503)

6) Unit tests added.


jasoncoolmax:
need to target to 2.3 branch

PR #652, opened 2016-05-26
Bug fixes in token cache and other source clean up

kpanwar:
B2C bug fixes contains
1) Rename variables to “token” instead of “accessToken”.
2) Add support to return access_token when returned from the service
3) Update cache storage to save only developer provided scopes.
4) Pass scope to token endpoint in interactive flows. This was needed
for cases where developer may get consent for scopes across multiple
resources. Token endpoint would not be able to determine without scopes
passed in to know how to scope the access/id token permissions (scp
claim).
5) Remove MRRT flag from authentication result.
6) Update the keychain query to use empty string for userId instead of
null to avoid “multiple entries found” error. We need this because in
some cases B2C developer may not have any userId to query the cache and
would need to pass empty string on his own.
7) Add profile as an API passed scope to the service.
8) renamed variables in test code
9) fixed a bug for expires_on and expires_in (matches dev branch now)
10) updated test data to account for fixes.


ADALiOS/ADALiOS/ADTokenCacheStoreItem+Internal.m
```diff
@@ -30,37 +30,38 @@ @implementation ADTokenCacheStoreItem (Internal)
 
 - (void)fillExpiration:(NSDictionary*)responseDictionary
 {
-    id expires_in = [responseDictionary objectForKey:@"expires_on"];
-    if (!expires_in)
+    id expires_in = [responseDictionary objectForKey:OAUTH2_EXPIRES_IN];
+    
+    //if access_token was not return then ONLY use id_token_expires_in
+    if (![NSString adIsStringNilOrBlank:[responseDictionary objectForKey:OAUTH2_ACCESS_TOKEN]])
```
RPangrle: I'm fairly sure that '!' shouldn't be at the front of the if check...

ADALiOS/ADALiOS/ADKeychainTokenCacheStore.m
```diff
@@ -153,7 +153,7 @@ + (BOOL)handleKeychainCode:(OSStatus)errCode
 - (void)logItem:(ADTokenCacheStoreItem*)item
```
RPangrle: There's a version number somewhere in here that should probably be incremented.

ADALiOS/ADALiOS/ADTokenCacheStoreItem+Internal.m
kpanwar: if there is no access token returned then only we use OAUTH2_ID_TOKEN_EXPIRES_IN (returned all the time) in the response. access_token is paired with expires_in and id_token is paired with id_token_expires_in. I cannot check if id_token exists because it is returned in all the responses. It is the lack of access token that matters.

RPangrle: This check will be true if there is an access token...

ADALiOS/ADALiOS/ADKeychainTokenCacheStore.m
RPangrle: I was referring to the token cache version here, just to prevent any issues that might pop up from reading older version cache.

ADALiOS/ADALiOS/ADALiOS.h
```diff
@@ -26,8 +26,8 @@
 #define STR_ADAL_VER_LOW    "0"
 #define STR_ADAL_VER_PATCH  "0"
 
-#define ADAL_VERSION_STRING     STR_ADAL_VER_HIGH "." STR_ADAL_VER_LOW "." STR_ADAL_VER_PATCH "-pre3"
```
RPangrle: pre4 has already been released, this should be pre5

ADALiOS/ADALiOS/ADTokenCacheStoreItem+Internal.m
kpanwar: fixed.

RPangrle:
:shipit:

kpanwar:
:shipit: for @jasoncoolmax's change

PR #663, opened 2016-06-09
Add app identifier prefix to keychain shared group

jasoncoolmax:
Application identifier prefix is now added when we set the keychain shared group.


jasoncoolmax:
@kpanwar

RPangrle:
:shipit:

Issue #666, opened 2016-06-14
ADAL does not present NTLM dialogue for ADFS/NTLM service


liwa:
CUSTOMER IMPACT- 
Majority of customer's iOS users are unable to sign into OneDrive or Yammer apps.
-PROBLEM- 
Customer's federation provider (Ping Federate) returns WWW-Authenticate: NTLM for users that do not have a kerberos profile installed on their device (majority of users). OneDrive will not prompt the user for NTLM credentials with this response.
Office Mobile apps prompt for NTLM credentials properly.
OneDrive uses ADAL 2.1.2, whereas Office Mobile is using 2.0.0.
This used to work a couple of months ago for OneDrive.

-Repro Steps-

The easiest way to reproduce the issue is with a standard IIS 401 response containing the WWW-Authenticate: NTLM header. Easily reproducible with Fiddler's AutoResponder:
1) From Fiddler >> AutoResponder tab >> Enable Rules.
2) Add a rule where:
a. If the request matches: https://msft.sts.microsoft.com/adfs/ls/?login_hint=
b. Respond with a text document containing the following:

HTTP/1.1 401 Unauthorized
Cache-Control: no-cache, no-store
Pragma: no-cache
Expires: -1
WWW-Authenticate: NTLM
X-Powered-By: ASP.NET
Date: Mon, 13 Jun 2016 15:00:55 GMT
Content-Length: 0
Proxy-Support: Session-Based-Authentication

4) Try signing into iOS OneDrive with your Microsoft Credentials while proxying through Fiddler. You will see a blank white screen which eventually times out.
a. Additionally, you can test signing into Word mobile with your Microsoft Credentials. You will see the NTLM credential prompt for Word, but not OneDrive.
-Investigation & Research- 
Office Mobile is still on an older build of ADAL (2.0.0).
OneDrive used to present this authentication box with older ADAL version.
OneDrive is now using ADAL 2.1.2, and no longer presenting the prompt.
Can repro using Fiddler. 
MICROSOFT PRODUCT VERSION AND SERVICE RELEASES: 
OneDrive 7.x
Yammer 6.6.8
ADAL 2.1.2, 2.2
HARDWARE/SOFTWARE ENVIRONMENT: 
iOS
-WORKAROUNDS (IF ANY)- 
None for this customer based on their authentication setup.

PR #668, opened 2016-06-17
Fix the bug where NTLM dialog does not show up

jasoncoolmax:
(for #666)
The NTLM dialog does not show up because, during redirect, the "ADURLProtocol" property in request is not properly set
Such a property is used to indicate which request will be intercepted and monitored for NTLM challenge:
if "ADURLProtocol" property is set, request will not be intercepted;
if "ADURLProtocol" property is NOT set, request will be intercepted.

Besides, fix another place that causes no-return when user click "cancel" button in NTLM dialog.

### DIFF
--- a/ADAL/src/urlprotocol/ADNTLMHandler.m
+++ b/ADAL/src/urlprotocol/ADNTLMHandler.m
@@ -98,7 +98,7 @@ static NSURLConnection *_conn = nil;
                 _challengeCancelled = YES;
                 AD_LOG_INFO_F(@"NTLM challenge cancelled", nil, @"host: %@", challenge.protectionSpace.host);
                 [challenge.sender performDefaultHandlingForAuthenticationChallenge:challenge];
-                [protocol stopLoading];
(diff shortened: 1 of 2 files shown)


ADAL/src/urlprotocol/ADNTLMHandler.m
```diff
@@ -98,7 +98,7 @@ + (BOOL)handleChallenge:(NSURLAuthenticationChallenge *)challenge
                 _challengeCancelled = YES;
                 AD_LOG_INFO_F(@"NTLM challenge cancelled", nil, @"host: %@", challenge.protectionSpace.host);
                 [challenge.sender performDefaultHandlingForAuthenticationChallenge:challenge];
-                [protocol stopLoading];
+                [protocol connection:connection didFailWithError:[NSError errorWithDomain:NSCocoaErrorDomain code:NSUserCancelledError userInfo:nil]];
             }
```
jasoncoolmax: without this line, webview won't return after user click "cancel" in NTLM dialog.

RPangrle:
:shipit:

PR #672, opened 2016-06-22
Move profile settings into ADTestAppSettings

RPangrle:
There was a bug before where we weren't properly getting/setting the auth profile on launch. Moving this functionality to ADTestAppSettings makes it easier to handle this case properly.


jasoncoolmax:
:shipit:

PR #673, opened 2016-06-22
Remove !TARGET_IPHONE_SIMULATOR ifdefs and keychain team ID change

RPangrle:
- I confirmed with an Apple Engineer at WWDC that Keychain Access Groups are intended to work in the iOS Simulator now in Xcode 7+
- The NSUserDefaults cache of the keychain ID could cause issues in older simulators, best to remove that and cache it on a per-app-launch basis instead. This also could have caused issues when apps were migrated between publishers.


jasoncoolmax:
:shipit:

PR #675, opened 2016-06-23
Remove resiliency from the public API

RPangrle:
Move the properties to turn on resiliency into the private API, this way it doesn't block us releasing.


jasoncoolmax:
I signed off :)

PR #678, opened 2016-06-27
WPJ Keychain Robustness

### LINKED ISSUE #671: Stop storing the keychain access group "hint" in NSUserDefaults
In -bundleSeedID, This is potentially problematic if an application gets migrated from one team ID to another. Instead using a singleton/dispatch_once to check this on first use.

RPangrle:
Add more robust error checking to retrieving the WPJ registration information, and bubble that information farther up the chain so we can act on it appropriately when the keychain is in an unexpected state.

Fixes #671 as well


kpanwar:
:shipit:

Issue #630, opened 2016-05-03
Fix header include paths when using ADAL as a static library


szalap:
When including ADAL through cocopoads, the ADAuthenticationContext.h complains about not having defined some classes like: WebViewType and ADAuthenticationCallback.
Manually adding the Adal.h import resolves the issue and lets the library together with project compile and build.

I saw that ADAL-prefix.pch contains ADAL_internal.h, but this does not seem to be imported.

RPangrle:
https://github.com/AzureAD/azure-activedirectory-library-for-objc/blob/master/ADAL/src/public/ADAL.h#L50

szalap:
It's other way around, it's ADAuthenticationContext.h, that does not see some definitions:
![screen shot 2016-05-04 at 14 34 59](https://cloud.githubusercontent.com/assets/2716955/15014295/8453694c-1205-11e6-8f2f-1c5124dc5286.png)

RPangrle:
How are you setting up your podfile?

szalap:
It's very basic, comes down to (pseudocode):
`target 'Project' do
pod 'ADAL', '~> 2.1'
...

post_install do | installer |
    require 'fileutils'
    post_install do | installer |
		require 'fileutils'
		FileUtils.cp_r('Pods/Target Support Files/Pods-Project/Pods-Project-acknowledgements.plist', 'project/Settings.bundle/Acknowledgements.plist', :remove_destination => true)
    end
end
`

szalap:
Seems like the `use_frameworks!` in Podfile is the answer.

PR #679, opened 2016-06-27
Setup the ADALiOS static library so the include path is properly

RPangrle:
So that it is ```<ADAL/ADAL.h>``` instead of ```<ADALiOS/ADAL.h>```  when consuming ADAL as a static library.

#630


jasoncoolmax:
:shipit:

PR #686, opened 2016-06-30
PKeyAuth No WPJ Unit Test

RPangrle:
Add a unit test that goes through the pkeyauth flow on the token endpoint when there's no WPJ. I didn't add other cases yet as anything that touches and uses keychain items becomes much more involved.


jasoncoolmax:
I signed off :)

Issue #698, opened 2016-07-15
The bundle seed ID entry should be marked always accessible


RPangrle:
This can cause failures if an app gets spun up in the background while the device is locked.

PR #699, opened 2016-07-15
Make keychainTeamId use an always accessible item

RPangrle:
The previous keychainTeamId check used an item which did not specify an accessibility attribute, which defaults to only being accessible while the device is unlocked. This can create issues for background applications that might kick up while the device is locked. Extra error checking and logging are being added on that keychain query as well to make sure we have a better idea what happened and to make sure ADAL doesn't try to do things which might make it worse.


jasoncoolmax:
:shipit:

RPangrle:
Fixes #698

Issue #702, opened 2016-07-19
ADNTLMHandler should not be using NSUserCancelledError


RPangrle:
We should avoid using system error codes for ADAL created errors, so that way we know if an error is coming from us or the OS. We already have AD_ERROR_UI_USER_CANCEL that can work here.

PR #703, opened 2016-07-19
Change ADNTLMHandler to return AD_ERROR_UI_USER_CANCEL on cancel

jasoncoolmax:
(for #702)

Changed the returning error code when NTLM dialog is canceled by user.


RPangrle:
:shipit:

PR #712, opened 2016-08-03
Extend ADAL to support WatchOS

ertang2015:
This commit include the changes to extend existing ADAL framework to work on WatchOS platform (Supporting WatchOS 2.0). There are several key changes to achieve the goal:

- A new ALDAL framework targeting at WatchOS. This include necessary source files to cover most ADAL functionality working properly on WatchOS. It doesn't support sign in interactively through ADAL as WatchOS doesn't support WebView. It doesn't support NTLM authentication either;

- Several new APIs to support WatchOS authenticate user by using existing authentication cache information stored on companion iOS. As long as iOS has authenticated user successfully, watchOS could ask for the stored cache and then calling the new APIs to authenticate locally on Watch;

- Replaced the NSURLConnection with NSURLSession on WatchOS since NSURLConnection is deprecated and forbidded on WatchOS.

### DIFF
--- a/ADAL/src/ADAL_Internal.h
+++ b/ADAL/src/ADAL_Internal.h
@@ -54,10 +54,12 @@
 typedef void(^ADAuthenticationCallback)(ADAuthenticationResult* result);
 
 
-#if TARGET_OS_IPHONE
+#if TARGET_OS_IOS
 //iOS:
 #   include <UIKit/UIKit.h>
 typedef UIWebView WebViewType;
+#elif TARGET_OS_WATCH
+typedef NSObject WebViewType; // Should never be called
 #else
 //OS X:
 #   include <WebKit/WebKit.h>
--- a/ADAL/src/ADAuthenticationContext.m
+++ b/ADAL/src/ADAuthenticationContext.m
@@ -346,6 +346,21 @@ NSString* ADAL_VERSION_VAR = @ADAL_VERSION_STRING;
     [request acquireToken:completionBlock];
 }
 
+#if TARGET_OS_WATCH
+- (void)acquireTokenSlientWithAuthData:(NSData *)authData
+                              resource:(NSString *)resource
+                              clientId:(NSString *)clientId
+                           redirectUri:(NSURL *)redirectUri
+                       completionBlock:(ADAuthenticationCallback)completionBlock
+{
+    API_ENTRY;
+    REQUEST_WITH_REDIRECT_URL(redirectUri, clientId, resource);
+    
+    [request setSilent:YES];
+    [request acquireTokenWithAuthData:authData completionBlock:completionBlock];
+}
+#endif
+
 - (void)acquireTokenSilentWithResource:(NSString*)resource
                               clientId:(NSString*)clientId
                            redirectUri:(NSURL*)redirectUri
--- a/ADAL/src/request/ADAuthenticationRequest+WebRequest.m
+++ b/ADAL/src/request/ADAuthenticationRequest+WebRequest.m
@@ -29,8 +29,10 @@
 #import "ADWebResponse.h"
 #import "ADPkeyAuthHelper.h"
 #import "ADAuthenticationSettings.h"
+#if !TARGET_OS_WATCH
 #import "ADWebAuthController.h"
 #import "ADWebAuthController+Internal.h"
+#endif
 #import "ADHelpers.h"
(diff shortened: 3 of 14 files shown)


ADAL/src/request/ADAuthenticationRequest+WebRequest.m
```diff
@@ -174,6 +176,14 @@ - (NSString*)generateQueryStringForRequestType:(NSString*)requestType
 - (void)launchWebView:(NSString*)startUrl
       completionBlock:(void (^)(ADAuthenticationError*, NSURL*))completionBlock
 {
+#if TARGET_OS_WATCH
+    (void)startUrl;
+    (void)completionBlock;
+    @throw [NSException exceptionWithName:@"WebViewException"
```
RPangrle: As a rule we don't throw exceptions in ADAL unless we otherwise would crash.

ADAL/src/ADAuthenticationContext.m
```diff
@@ -346,6 +346,21 @@ - (void)acquireTokenWithResource:(NSString*)resource
...
+                              resource:(NSString *)resource
+                              clientId:(NSString *)clientId
+                           redirectUri:(NSURL *)redirectUri
+                       completionBlock:(ADAuthenticationCallback)completionBlock
+{
+    API_ENTRY;
+    REQUEST_WITH_REDIRECT_URL(redirectUri, clientId, resource);
+    
+    [request setSilent:YES];
+    [request acquireTokenWithAuthData:authData completionBlock:completionBlock];
```
RPangrle: This isn't going to work if tokens ever become device bound.

ADAL/src/ADAL_Internal.h
```diff
@@ -54,10 +54,12 @@
 typedef void(^ADAuthenticationCallback)(ADAuthenticationResult* result);
 
 
-#if TARGET_OS_IPHONE
+#if TARGET_OS_IOS
 //iOS:
 #   include <UIKit/UIKit.h>
 typedef UIWebView WebViewType;
+#elif TARGET_OS_WATCH
```
RPangrle: If would rather see any APIs that used webviews ifdef'd out for watch devices

ADAL/src/request/ADWebRequest+NSURLSession.h
```diff
@@ -0,0 +1,71 @@
...
+@class ADWebRequest;
+@class ADWebResponse;
+
+typedef enum
+{
+    ADWebRequestGet,
+    ADWebRequestPost,
+} ADWebRequestMethodType;
+
+@interface ADWebRequest : NSObject
```
RPangrle: Divergent implementations of ADWebRequest within the same code base should not exist like this.

ADAL/src/ADAuthenticationContext.m
ertang2015: May I ask when will the token be device bounded? Are there any specific code handling device bound tokens?

ADAL/src/request/ADWebRequest+NSURLSession.h
ertang2015: Are you suggesting refactoring the code using Category?

ADAL/src/ADAuthenticationContext.m
RPangrle: It's a feature we expect to add in the future, and RTs should not be moved between devices. What should happen is the phone refreshes the token, if necessary, and passes the updated AT to the watch.

ertang2015: Thanks for the clarification. The reason we are doing this work is trying to make sure a watch app can work *standalone* as much as possible. This means, once the watch get the initial authentication info from Phone, it should be able to refresh by itself without the help of the phone. That's why we are trying to pass the refresh token from phone to watch.

ADAL/src/request/ADWebRequest+NSURLSession.h
RPangrle: Well, part of the problem is that you named this file as if it is a category but it is not a category at all. It's a wholesale separate implementation of ADWebRequest.  There should not be separate implementations of the same class except for either thin shims or UI. If your requirement here is to use NSURLSession then all ADWebRequests should be moved to NSURLSession, and not create extra technical debt in ADAL. As it stands here you've created a complete separate code path that has no test coverage at all.

PR #713, opened 2016-08-03
Fix iOS 10 issue on 1.2.x

RPangrle:
Remove the calls to didFailWithError that are causing iOS 10 Beta 2+ to fail


jasoncoolmax:
:shipit:

Issue #694, opened 2016-07-07
ADAL framework require iOS 9.2 ?


eddiekim:
The ADAL framework target has a deployment target explicitly set to 9.2. Is that a requirement? Our app needs to support 8.0+

![adal-deployment-target](https://cloud.githubusercontent.com/assets/1395025/16664855/e6f065a6-4436-11e6-9db9-fdcdddff8f8c.png)

RPangrle:
The static lib itself still has a deployment target of iOS 7 (even though our support policy says iOS 8)...

Issue #97, opened 2014-06-12
Add an Swift app example using AAD


greathansen:
Could you provide and example off a swift app using the existing AAD library?

regards,
Gustavo

nimatra:
Please don't won't fix these requests. Every ADAL newbie needs to go through the same path. Just one simple example would actually benefit ADAL team more. 
We (startup devs) would just change our mind and use Google/FB instead of ADAL if it's hard to use.

RandalliLama:
@nimatra I'll reopen this.  We have nothing against Swift, we are just swamped right now.  We would happily accept a community contribution if you have one.

nimatra:
I started learning/doing Swift and AD B2C yesterday. I'll happily post a sample as soon as I get my AD running on iPhone.

RandalliLama:
@nimatra Got it.  Did you happen to look at the sample I posted above?  It's not pretty, but it was relatively simple to get it working.  Note that the sample was done with an earlier version of Swift.

May I also suggest posting a question to StackOverflow to see if anyone else has already done this?

nimatra:
I found some clues on SO and bridge header helps.
Generally, having a working sample project that I can just replace the AD info with mine from my other apps and see it working on iOS helps a ton.

tsteele2013:
Is a working Swift example still needed?

OceanOSdev:
A working swift example would be much appreciated.

tsteele2013:
Inside my AuthManager class is the following:


    let tenant:String = "TENANT.onmicrosoft.com"
    let authority:String = "https://login.windows.net/TENANT.onmicrosoft.com"
    let clientID:String = "YOUR CLIENT ID"
    let resourceURI:String = "YOUR RESOURCE URI"
    let redirectURI:NSURL = NSURL(string:"YOUR REDIRECT URI")!
    var loggedIn: Bool = false
    // Create a singleton instance of the AuthManager class. The app uses this instance for all operations.
    class var sharedInstance: AuthManager {
        struct Singleton {
            static let instance = AuthManager()
        }
        return Singleton.instance
    }
    
    func login(completionBlock:((Bool, String) -> Void)) {
                var er:ADAuthenticationError? = nil
        let authContext = ADAuthenticationContext(authority: authority, error: &er)
        var token: String!
[...]

Issue #693, opened 2016-07-06
Importing ADAL module results in build error (v 2.2.1)


eddiekim:
I get build errors on v2.2.1 when semantically importing the ADAL module (`import ADAL`). This was fine in v2.2.0, and appears to be related to commit https://github.com/AzureAD/azure-activedirectory-library-for-objc/commit/91d7000b8c6eefc2846f700657c9bc624a2ecec0 . Seems that Xcode doesn't like that the same header files are made available by both the `ADALiOS` static library and `ADAL` framework by the same path.

![adal-module-error](https://cloud.githubusercontent.com/assets/1395025/16638139/af59f802-4399-11e6-884b-83d50aacef3a.png)

To reproduce in `MyTestiOSApp`:

* Remove `libADALiOS.a` from `Linked Frameworks and Libraries`
* Add `ADAL.framework` to `Embedded Binaries`
* Replace `#import <ADAL/ADAL.h>` with `@import ADAL;` in `ADTestAppDelegate.m`
* **Result**: `error: include of non-modular header inside framework module 'ADAL' `
* Workaround: go to `Build Phases` for `ADALiOS` static library and change `Copy Files` `Subpath` to something else (anything)
* Clear derived data and build again
* **Result**: Build success

RPangrle:
Unfortunately that change was made so that applications only using the static libraries can import headers properly. The only way I can think around this is to not have the ADAL framework rely on the ADALiOS static lib

eddiekim:
I think you will have more people who prefer framework to static libs going forward, so it's something to think about. As-is, it doesn't seem like you can use the `ADAL` framework unless you build it outside of the client project first (or make changes to the Xcode project)

This isn't really an issue for us anymore, so feel free to close.

RPangrle:
I've done some work to mitigate this. The static lib build has been split into two libs, one that'll copy headers and one that won't. This way you'll be able to include the xcode project in your project and choose framework or static libs, you won't be able to do both though.

PR #714, opened 2016-08-05
Better Builds, Extension Safe and Swift Checking

RPangrle:
Pushing some improvements that have been sitting on my dev box for a little while.

* An improved build script.
There's a single build script written in Python now that builds all of our libraries and frameworks and test apps. It also spits out better output and a summary at the end so it should be easier to run a validation build on your dev box before check in.

* Sample Swift App
Added a sample swift app that consumes are framework to make sure that build doesn't get broken and provide a simple bit of sample code on how to use the API in Swift.

* Cleaned up Extension Build Paths
Separated out the "core" libraries from targets that copy headers so they don't conflict with framework includes, however if you build both in the same dev environment it still will break things as the underlying issue is a bug with how Xcode prioritizes header include paths vs framework include paths.

* Test Share Extension added to the Test App
Doesn't test much except the build right now, but is marginally useful for future test code.


jasoncoolmax:
:shipit:

RPangrle:
Fixed #694 deployment target set to 8.0 across the board now.

RPangrle:
Mitigates #693

RPangrle:
Fixed #97

Issue #710, opened 2016-08-02
Thread safety issues in -[ADTokenCache serialize]


atdrendel:
We are seeing relatively frequent crashes for some of our users on 10.12. We are using ADAL 2.2.2. The crashes seem to be internal `NSKeyedArchiver` crashes. Is this known?

atdrendel:
```
Thread 9 Crashed:
0   libobjc.A.dylib                      0x0000000107a8aa31 objc_retain + 33
1   CoreFoundation                       0x00000001096aae11 __CFBasicHashAddValue + 1777
2   CoreFoundation                       0x00000001096ab89d CFDictionarySetValue + 221
3   Foundation                           0x000000010746e936 _encodeObject + 645
4   ADAL                                 0x0000000113d0d493 -[ADTokenCacheItem encodeWithCoder:] (ADTokenCacheItem.m:151)
5   Foundation                           0x000000010746eb8a _encodeObject + 1241
6   Foundation                           0x000000010747013c -[NSKeyedArchiver _encodeArrayOfObjects:forKey:] + 460
7   Foundation                           0x0000000107481312 -[NSDictionary(NSDictionary) encodeWithCoder:] + 927
8   Foundation                           0x000000010746eb8a _encodeObject + 1241
9   Foundation                           0x000000010747013c -[NSKeyedArchiver _encodeArrayOfObjects:forKey:] + 460
10  Foundation                           0x0000000107481312 -[NSDictionary(NSDictionary) encodeWithCoder:] + 927
11  Foundation                           0x000000010746eb8a _encodeObject + 1241
12  Foundation                           0x000000010747013c -[NSKeyedArchiver _encodeArrayOfObjects:forKey:] + 460
13  Foundation                           0x0000000107481312 -[NSDictionary(NSDictionary) encodeWithCoder:] + 927
14  Foundation                           0x000000010746eb8a _encodeObject + 1241
15  Foundation                           0x000000010747013c -[NSKeyedArchiver _encodeArrayOfObjects:forKey:] + 460
16  Foundation                           0x0000000107481312 -[NSDictionary(NSDictionary) encodeWithCoder:] + 927
17  Foundation                           0x000000010746eb8a _encodeObject + 1241
18  Foundation                           0x00000001074aa5a2 +[NSKeyedArchiver archivedDataWithRootObject:] + 156
[...]

PR #721, opened 2016-08-12
Fix thread safety issues in -[ADTokenCache serialize]

atdrendel:
ADTokenCache accesses its _cache instance variable, which is an NSMutableDictionary, in an unsafe manner. This pull request naively synchronizes all access to the _cache. See https://github.com/AzureAD/azure-activedirectory-library-for-objc/issues/710

### DIFF
--- a/ADAL/src/cache/ADTokenCache.m
+++ b/ADAL/src/cache/ADTokenCache.m
@@ -71,10 +71,12 @@
 
 - (id)cacheObjectForKey:(id<NSCopying>)key
 {
+    id object = nil;
     @synchronized(self)
     {
-        return _cache[key];
+        object = _cache[key];
     }
+    return object;
 }
 
 - (void)setCache:(NSMutableDictionary *)newCache
@@ -89,10 +91,13 @@
 
 - (NSMutableDictionary *)cacheCopy
 {
+    NSMutableDictionary *copy = nil;
     @synchronized(self)
     {
-        return [_cache mutableCopy];
+        copy = [_cache mutableCopy];
     }
+    SAFE_ARC_AUTORELEASE(copy);
+    return copy;
 }
 
 - (BOOL)hasCache
--- a/ADAL/src/cache/ADTokenCache.m
+++ b/ADAL/src/cache/ADTokenCache.m
@@ -69,6 +69,40 @@
     SAFE_ARC_SUPER_DEALLOC();
 }
 
+- (id)cacheObjectForKey:(id<NSCopying>)key
+{
+    @synchronized(self)
+    {
+        return _cache[key];
+    }
+}
+
+- (void)setCache:(NSMutableDictionary *)newCache
+{
+    @synchronized(self)
+    {
+        SAFE_ARC_RELEASE(_cache);
+        _cache = newCache;
+        SAFE_ARC_RETAIN(_cache);
+    }
+}
+
+- (NSMutableDictionary *)cacheCopy
+{
+    @synchronized(self)
+    {
+        return [_cache mutableCopy];
+    }
+}
+
+- (BOOL)hasCache
+{
+    @synchronized(self)
+    {
+        return _cache != nil;
+    }
+}
+
 - (void)setDelegate:(nullable id<ADTokenCacheDelegate>)delegate
 {
     if (_delegate == delegate)
--- a/ADAL/src/cache/ADTokenCache.m
+++ b/ADAL/src/cache/ADTokenCache.m
@@ -112,8 +112,7 @@
     SAFE_ARC_RELEASE(_delegate);
     _delegate = delegate;
     SAFE_ARC_RETAIN(_delegate);
-    SAFE_ARC_RELEASE(_cache);
-    _cache = nil;
+    [self setCache:nil];
     
     if (!delegate)
     {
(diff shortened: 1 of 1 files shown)


ADAL/src/cache/ADTokenCache.m
```diff
@@ -567,7 +603,7 @@ - (BOOL)addOrUpdateImpl:(ADTokenCacheItem *)item
 
 - (NSString *)description
 {
-    return [NSString stringWithFormat:@"ADTokenCache: %@", _cache];
+    return [NSString stringWithFormat:@"ADTokenCache: %@", [self cacheCopy]];
```
RPangrle: It would be better to grab the lock then create a copy here (which you're going to have to grab to create the copy in the first place...)

ADAL/src/cache/ADTokenCache.m
```diff
@@ -69,6 +69,45 @@ - (void)dealloc
...
+}
+
+- (NSMutableDictionary *)cacheCopy
+{
+    NSMutableDictionary *copy = nil;
+    @synchronized(self)
+    {
+        copy = [_cache mutableCopy];
+    }
+    SAFE_ARC_AUTORELEASE(copy);
```
RPangrle: According to the Create Rule you should own anything with "copy" in the method name:

https://developer.apple.com/library/ios/documentation/CoreFoundation/Conceptual/CFMemoryMgmt/Concepts/Ownership.html

RPangrle: Wait... "starts with" I still don't like it as it's unnecessarily confusing (kind of like Apple's slight differences between the CF and ObjC create rules...)

ADAL/src/cache/ADTokenCache.m
```diff
@@ -69,6 +69,45 @@ - (void)dealloc
     SAFE_ARC_SUPER_DEALLOC();
 }
 
+- (id)cacheObjectForKey:(id<NSCopying>)key
+{
+    id object = nil;
+    @synchronized(self)
```
RPangrle: The usage pattern for ADAL cache tends to be fairly read heavy, as such I would like to avoid mutexes and these small methods.

ADAL/src/cache/ADTokenCache.m
atdrendel: I don't understand. -cacheCopy returns an autoreleased copy of the _cache object, which should be completely safe to use here. I'd hate to have to add a @synchronize call here if I don't have to.

RPangrle: Creating a copy just to display it in a description is a tad wasteful, and creating that copy required you to use @synchronize anyways. And creating that copy will take longer then it took create the string...

ADAL/src/cache/ADTokenCache.m
atdrendel: Yeah, "newObject", "copyObject", or "initObject" return owned objects. So, this method name should be safe. The broader point (that including the word "copy" in the method name is weird) still stands, though. I'm not really sure how to get around it unless we want to pollute this method with a bunch of `@synchronize`s.

ADAL/src/cache/ADTokenCache.m
RPangrle: I'm right now in the middle of taking a crack at this with r/w locks. Like I said the usage pattern of the ADAL cache is extremely read heavy so I'd like to avoid mutexes.

ADAL/src/cache/ADTokenCache.m
atdrendel: I could certainly add the `@synchronize` if you want. I tend to prefer readability, especially in a method like `description` which is rarely called and should never be a performance problem. It's up to you, however.

ADAL/src/cache/ADTokenCache.m
atdrendel: What's the alternative? I could add the `@synchronize` calls inline, but it doesn't remove the need for the mutex. We'd still need to add the `@synchronize`s, unless you have a different strategy?

ADAL/src/cache/ADTokenCache.m
RPangrle: As a general rule making extra copies of things should be avoided unless making that copy is going to be faster then what you're trying to do. In -serialize that's probably the case, most everywhere else probably not.

ADAL/src/cache/ADTokenCache.m
RPangrle: pthread_rwlock. It's kinda gnarly but more performant :-/

atdrendel: pthread-rwlock isn't re-entrant, is it? `getItemsWithKey:userId:correlationId:error:` calls `getItemsImplKey:userId:`.

atdrendel: Never mind. It's only called in that method. So, `getItemsImplKey:userId:` doesn't require a lock in the method itself.

RPangrle: Yeah. Most of these are split between an outwardly visible method and an impl method. Take a look: https://github.com/AzureAD/azure-activedirectory-library-for-objc/pull/722

atdrendel: Looks good. How long do pull requests usually take to get merged?

RPangrle: As long as it takes me to hit the 'merge' button. Getting out a release is another matter though.

PR #722, opened 2016-08-12
Add a rwlock to protect the _cache dictionary

RPangrle:
Bad things can happen if the cache dictionary gets modified while we're trying to serialize it. Protect it with a rw lock.


kpanwar:
:shipit:

RPangrle:
Fixes #710

PR #723, opened 2016-08-13
whitelist redirect uri that starts with "http://"

### LINKED ISSUE #720: Handle redirect uri that starts with "http://"
For ADURLProtocol, we need to make sure the redirect uri, which is the end url used to go back to the app, goes through even though it is in the form of "http://***".

jasoncoolmax:
Fix #720 

Url loaded in webview will be filtered out if it is a http connection. Now changes have been made to whitelist the redirect uri, which is the end url used to go back to the app, regardless of its uri format.

### DIFF
--- a/ADAL/src/urlprotocol/ADURLProtocol.m
+++ b/ADAL/src/urlprotocol/ADURLProtocol.m
@@ -62,8 +62,11 @@ static NSUUID * _reqCorId(NSURLRequest* request)
 
 + (BOOL)registerProtocol:(NSString*)endURL
 {
-    s_endURL = endURL;
-    SAFE_ARC_RETAIN(s_endURL);
+    if (s_endURL!=endURL)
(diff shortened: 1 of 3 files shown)


RPangrle:
:shipit:

ADAL/src/urlprotocol/ADURLProtocol.m
```diff
@@ -59,14 +60,18 @@ + (void)registerHandler:(id)handler
 }
 
 
-+ (BOOL)registerProtocol
++ (BOOL)registerProtocol:(NSString*)endURL
 {
+    s_endURL = endURL;
```
RPangrle: Check to make sure s_endURL != endURL first. Shouldn't happen in practice, but it's best to be safe.

RPangrle:
:shipit:

PR #725, opened 2016-08-18
Fixed -25299 failure in retrieveTeamIDFromKeychain when using the Intune MAM SDK

msintuneappsdk:
For compatibility with the Intune MAM SDK, the SecItemCopyMatching query cannot contain the  kSecAttrAccessible attribute.


ADAL/src/utils/ios/ADKeychainUtil.m
```diff
@@ -53,15 +53,16 @@ + (NSString*)retrieveTeamIDFromKeychain:(ADAuthenticationError * __autoreleasing
     NSDictionary *query = @{ (id)kSecClass : (id)kSecClassGenericPassword,
                              (id)kSecAttrAccount : @"teamIDHint",
```
RPangrle: We renamed this item. In older SDKs it was "bundleSeedID" check your code to see what else might be adding "teamIDHint"

ADAL/src/utils/ios/ADKeychainUtil.m
```diff
@@ -53,15 +53,16 @@ + (NSString*)retrieveTeamIDFromKeychain:(ADAuthenticationError * __autoreleasing
...
                              (id)kSecReturnAttributes : @YES };
     CFDictionaryRef result = nil;
     
     OSStatus status = SecItemCopyMatching((__bridge CFDictionaryRef)query, (CFTypeRef *)&result);
     
     if (status == errSecItemNotFound)
     {
-        status = SecItemAdd((__bridge CFDictionaryRef)query, (CFTypeRef *)&result);
+        NSMutableDictionary* addQuery = [query mutableCopy];
+        [addQuery setObject:(id)kSecAttrAccessibleAlways forKey:(id)kSecAttrAccessible];
```
RPangrle: This does not address the original bug where if the item is still around with the wrong accessibility attribute if the device was locked we wouldn't be able to get the keychain team ID.

ADAL/src/utils/ios/ADKeychainUtil.m
msintuneappsdk: Changed the PR description to better reflect the issue. The Intune SDK DLP changes the kSecAttrAccessible attribute from kSecAttrAccessibleAlways to kSecAttrAccessibleAlwaysThisDeviceOnly in the original SecItemAdd call. Since kSecAttrAccessibleAlways is included in the SecItemCopyMatching query, applications not managed by the Intune SDK will fail to find the keychain item and will attempt to add it, resulting in a -25299 error. The kSecAttrAccessible attribute is not needed in the SecItemCopyMatching query. It's only needed in the SecItemAdd query.

ADAL/src/utils/ios/ADKeychainUtil.m
msintuneappsdk: This change still creates the item using the kSecAttrAccessibleAlways attribute. However, it removes it from the SecItemCopyMatching query which will allow it to be found regardless of its kSecAttrAccessible value which will be kSecAttrAccessibleAlwaysThisDeviceOnly if it was created by an Intune MAM managed application.

RPangrle: If that's the case shouldn't we just be setting it to kSecAttrAccessibleAlwaysThisDeviceOnly in all cases then?

msintuneappsdk: That wouldn't hurt, but the same issue would be hit for current 2.2.2 apps. There could be devices with the teamIDHint value already created with kSecAttrAccessibleAlways. If the SecItemCopyMatching query contained kSecAttrAccessible==kSecAttrAccessibleAlwaysThisDeviceOnly, it wouldn't be able to find the old items with kSecAttrAccessible==kSecAttrAccessibleAlways. The safest change would be to leave the kSecAttrAccessible attribute out of the SecItemCopyMatching query and only put it in from the SecItemAdd query.

PR #726, opened 2016-08-19
Fix issue in ADURLProtocol when URLs get canonicalized

RPangrle:
Remove  [self.client URLProtocol:self wasRedirectedToRequest:mutableRequest redirectResponse:response] call. This would cause multiple requests to the same endpoint that could break some configurations.

Replace the @"ADURLProtocol" string with a constant.

### DIFF
--- a/ADAL/src/ui/ADWebAuthController.m
+++ b/ADAL/src/ui/ADWebAuthController.m
@@ -417,6 +417,19 @@ NSString* ADWebAuthWillSwitchToBrokerApp = @"ADWebAuthWillSwitchToBrokerApp";
     {
         return;
     }
+    
+    // If we failed on an invalid URL check to see if it matches our end URL
+    if ([error.domain isEqualToString:@"NSURLErrorDomain"] && error.code == -1002)
+    {
+        NSURL* url = [error.userInfo objectForKey:NSURLErrorFailingURLErrorKey];
+        if ([[[url absoluteString] lowercaseString] hasPrefix:_endURL])
+        {
+            _complete = YES;
+            [self webAuthDidCompleteWithURL:url];
+            return;
+        }
+        
+    }
 
     if (error)
     {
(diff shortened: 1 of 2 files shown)


ADAL/src/ui/ADWebAuthController.m
```diff
@@ -417,6 +417,19 @@ - (void)webAuthDidFailWithError:(NSError *)error
     {
         return;
     }
+    
+    // If we failed on an invalid URL check to see if it matches our end URL
+    if ([error.domain isEqualToString:@"NSURLErrorDomain"] && error.code == -1002)
+    {
+        NSURL* url = [error.userInfo objectForKey:NSURLErrorFailingURLErrorKey];
+        if ([[[url absoluteString] lowercaseString] hasPrefix:_endURL])
```
jasoncoolmax: should be [_endURL lowercaseString]? :)

ADAL/src/urlprotocol/ADURLProtocol.m
```diff
@@ -202,36 +213,38 @@ - (void)connection:(NSURLConnection *)connection
...
     (void)connection;
     
-    AD_LOG_VERBOSE_F(@"-[ADURLProtocol connection:willSendRequest:]", _correlationId, @"Redirect response: %@. New request:%@", response.URL, request.URL);
+    AD_LOG_VERBOSE_F(@"-[ADURLProtocol connection:willSendRequest:]", _correlationId, @"Redirect response: %@. New request:%@", redirectResponse.URL, request.URL);
     
     // Disallow HTTP for ADURLProtocol
     if ([request.URL.scheme isEqualToString:@"http"])
     {
-        if (!s_endURL)
+        if ([request.URL.absoluteString.lowercaseString hasPrefix:s_endURL])
```
jasoncoolmax: Seems here it is only targetting at end URL in the form of "http://****", 
are we supposed to return error for all kinds of end url? :P

RPangrle: non http/https can't be loaded in Safari anyways, and https URLs don't have the same security implications http ones do.

jasoncoolmax:
:shipit:

PR #731, opened 2016-08-23
Fix logging error from UI code copy-over

RPangrle:
Remove an errant 'nil' that should have been removed in the copy-and-paste and guard against any nils that might get passed through to the logged in the future (just in case)


jasoncoolmax:
:shipit:

PR #738, opened 2016-08-26
Output Single iOS Framework Distribution

joeydong:
This switches the strategy for handling iOS application extension hosts to using a runtime check instead of providing a separate framework output. See commit messages for details.

We still have a bunch of validation to do before this is suitable for merging but I'm opening the PR for visibility.

By the way, have we considered using a mocking framework to simplify the unit test writing?


ADAL/src/ui/ios/UIAlertView+Additions.m
```diff
@@ -1,51 +1,63 @@
...
+            bundle = [NSBundle mainBundle];
+        }
+
+        alert = [[UIAlertView alloc] initWithFrame:CGRectZero];
+        alert.title = NSLocalizedStringFromTableInBundle(@"Enter your credentials", nil, bundle, nil);
+        alert.alertViewStyle = UIAlertViewStyleLoginAndPasswordInput;
+        
+        [alert addButtonWithTitle:NSLocalizedStringFromTableInBundle(@"Cancel", nil, bundle, nil)];
+        [alert addButtonWithTitle:NSLocalizedStringFromTableInBundle(@"Login", nil, bundle, nil)];
+        [alert setCancelButtonIndex:0];
```
joeydong: I'm not totally confident about this change. How do we reproduce this flow?

joeydong:
I'm not sure how the frameworks are getting imported in the sample project. How do I fix that?

ADAL/src/utils/ios/ADAppExtensionUtil.m
```diff
@@ -0,0 +1,68 @@
...
+
+@implementation ADAppExtensionUtil
+
++ (BOOL)isExecutingInAppExtension
+{
+    NSString* mainBundlePath = [[NSBundle mainBundle] bundlePath];
+
+    if (mainBundlePath.length == 0)
+    {
+        AD_LOG_ERROR(@"Expected `[[NSBundle mainBundle] bundlePath]` to be non-nil. Defaulting to non-application-extension safe API.", AD_ERROR_UNEXPECTED, nil, nil);
```
eddiekim: Not sure that this should be an error since this path will be part of the normal flow from an extension.

joeydong: The `[NSBundle mainBundle]` returns something in both cases:

- When the host is an app, the `bundlePath` is something like `file://.../app-ios.app`
- When the host is an extension, the `bundlePath` is something like `file://.../today-widget.appex`

ADAL/src/ui/ios/UIAlertView+Additions.m
RPangrle: You need to get a server to send you an NTLM challenge. The only way I have to do it involves using a MS STS while on corpnet and your corpnet account. I'll try to repro and send you directions.

RPangrle:
@joeydong, the SampleSwiftApp scheme itself is set to build the ADAL framework first and then the Sample Swift App. Considering that error doesn't repro cleanly I'm wondering if xcodebuild is getting confused from some other state laying around.

ADAL/src/utils/ios/ADAppExtensionUtil.m
eddiekim: oh sorry, I misunderstood. never mind

joeydong:
thx for reviewing @RPangrle / @eddiekim

PR #739, opened 2016-08-26
Fixed small typos in README

adamwulf:
Found these small typos when reading through the README while I was integrating into a Mac app. Integration has been smooth :) These small fixes are the least I can do


joeydong:
thanks @adamwulf !

Issue #740, opened 2016-08-29
Generate the various files with version info as part of the build


RPangrle:
Right now there is 4 different places in ADAL_Internal.h that have version information in them, on top of the info.plist files for the various ADAL frameworks. There should only be one place where this version information is defined and then generate the rest in a build script.

PR #744, opened 2016-08-31
Add build script to generate version number for info.plist

jasoncoolmax:
(address #740)

- A build script has been added, which will read version number from ADAL_Internal.h and add it to each Info.plist file.
(In order to run this script before an Info.plist is processed, i have to create a new build target called "ADAL Preprocessing", and then make other target dependent on it.)

- Changes have been made to the Macro in ADAL_Internal.h, such that version nubmers are defined in a single place.

### DIFF
--- a/ADAL/ADAL.xcodeproj/project.pbxproj
+++ b/ADAL/ADAL.xcodeproj/project.pbxproj
@@ -269,6 +269,27 @@
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
+		60967E131D76B5CC00863853 /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 8B0965A217F25770002BDFB8 /* Project object */;
+			proxyType = 1;
+			remoteGlobalIDString = 60967E071D76B56D00863853;
+			remoteInfo = "ADAL Preprocessing";
+		};
+		60967E151D76B5D300863853 /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 8B0965A217F25770002BDFB8 /* Project object */;
+			proxyType = 1;
+			remoteGlobalIDString = 60967E071D76B56D00863853;
+			remoteInfo = "ADAL Preprocessing";
+		};
+		60967E171D76B5E000863853 /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 8B0965A217F25770002BDFB8 /* Project object */;
+			proxyType = 1;
+			remoteGlobalIDString = 60967E071D76B56D00863853;
+			remoteInfo = "ADAL Preprocessing";
+		};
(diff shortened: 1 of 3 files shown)


ADAL/ADAL.xcodeproj/project.pbxproj
```diff
@@ -1311,7 +1377,30 @@
...
+			buildActionMask = 2147483647;
+			files = (
+			);
+			inputPaths = (
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "/usr/bin/python ${PROJECT_DIR}/tools/update_build_version.py ${PROJECT_DIR}/src/ADAL_Internal.h ${PROJECT_DIR}/resources/ios/Framework/Info.plist ${PROJECT_DIR}/resources/ios/ADALiOSBundle/ADALiOSBundle-Info.plist";
```
jasoncoolmax: We need to put all the Info.plist files as parameters for the above command, because they are in different paths.

Is it better to put all the Info.plist files in a single folder, such that the build script can locate all the Info.plist files?

RPangrle: The infoplist file path is available in a build variable, you should be able to use that in your script.

ADAL/ADAL.xcodeproj/project.pbxproj
```diff
@@ -1127,6 +1168,24 @@
 /* End PBXHeadersBuildPhase section */
 
 /* Begin PBXNativeTarget section */
+		60967E071D76B56D00863853 /* ADAL Preprocessing */ = {
```
RPangrle: Avoid having a separate target for this as it muddles up the dependencies, it'd be better to just add the run script phase in each of the applicable target

jasoncoolmax: I notice in the build log that, "Process Info.plist" always happens before the Run Script, even after I put Run Script as top as I can in the "Build Phase" setting.

But I will try to dig deeper to see whether there is a way to do it :)

RPangrle: Apple does some of their own processing that we might be able to use, especially if we provide the build version as a build variable in an xcconfig file.

ADAL/ADAL.xcodeproj/project.pbxproj
jasoncoolmax: Yep, now using ${BUILT_PRODUCTS_DIR}/${INFOPLIST_PATH} :)

ADAL/ADAL.xcodeproj/project.pbxproj
jasoncoolmax: Now it works by running script for each target :)

I find that I should actually set the version number in ${BUILT_PRODUCTS_DIR}/${INFOPLIST_PATH}.

RPangrle:
:shipit:

PR #747, opened 2016-09-07
NTLM fix

kpanwar:
this PR includes fixes for supporting NTLM auth and ensuring the webview is closed when the user presses cancel on the NTLM dialog.


RandalliLama:
:shipit:

PR #748, opened 2016-09-07
Bring back the connection canceling and error sending. 

RPangrle:
Trap the error in the webview delegate.

Canceling out the connection solves the double-connection issue while still allowing NTLM to work.


jasoncoolmax:
I signed off.

PR #750, opened 2016-09-08
Added nullable indicators to ADTokenCacheItem to enable Swift support

clausjoergensen:
Resource can be nil as it is specified in the comment therefore it makes sense to add "nullable" in Objective-C layer, which will generate optimal String? type in Swift projection.

### DIFF
--- a/ADAL/src/public/ADTokenCacheItem.h
+++ b/ADAL/src/public/ADTokenCacheItem.h
@@ -78,7 +78,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (copy, nullable) NSDate* expiresOn;
 
 
-@property (retain) ADUserInformation* userInformation;
+@property (retain, nullable) ADUserInformation* userInformation;
 
 /*!
  The item is a tombstone if this property if not nil;
--- a/ADAL/src/public/ADTokenCacheItem.h
+++ b/ADAL/src/public/ADTokenCacheItem.h
@@ -55,8 +55,10 @@
     NSMutableDictionary* _additionalClient;
 }
 
+NS_ASSUME_NONNULL_BEGIN
+
 /*! Applicable resource. Should be nil, in case the item stores multi-resource refresh token. */
-@property (copy) NSString* resource;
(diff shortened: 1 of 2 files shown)


ADAL/src/public/ADTokenCacheItem.h
```diff
@@ -65,15 +67,15 @@
...
-@property (copy) NSString* refreshToken;
+@property (copy, nullable) NSString* refreshToken;
 
 @property (copy) NSData* sessionKey;
 
-@property (copy) NSDate* expiresOn;
+@property (copy, nullable) NSDate* expiresOn;
 
 
 @property (retain) ADUserInformation* userInformation;
```
RPangrle: userInformation is also nullable

clausjoergensen: Added it to userInformation as well

ADAL/src/public/ADTokenCacheItem.h
```diff
@@ -65,18 +67,18 @@
...
 /*! The access token received. Should be nil, in case the item stores multi-resource refresh token. */
-@property (copy) NSString* accessToken;
+@property (copy, nullable) NSString* accessToken;
 
 @property (copy) NSString* accessTokenType;
 
-@property (copy) NSString* refreshToken;
+@property (copy, nullable) NSString* refreshToken;
 
 @property (copy) NSData* sessionKey;
```
RPangrle: sessionKey is also nullable, and usually is null. in fact the only thing in here that absolutely can't be null is authority.

ADAL/src/public/ADTokenCacheItem.h
```diff
@@ -55,28 +55,30 @@
     NSMutableDictionary* _additionalClient;
 }
 
+NS_ASSUME_NONNULL_BEGIN
```
RPangrle: Sorry for the back-and-forth here, wouldn't it be best to remove this line as well? Force devs to define the nullability if we add anything instead of having an assumption foisted on them and they weren't aware of.

PR #758, opened 2016-09-10
Copy over iOS 10/NTLM/Shib fixes to 1.2.x

RPangrle:


ADALiOS/ADALiOS/ADURLProtocol.m
```diff
@@ -102,29 +119,66 @@ -(void) connection:(NSURLConnection *)connection
...
+        if ([request.URL.absoluteString.lowercaseString hasPrefix:s_endURL])
+        {
+            // In this case we want to create an NSURLError so we can intercept the URL in the webview
+            // delegate, while still forcing the connection to cancel. This error is the same one the
+            // OS sends if it's unable to connect to the host
+            [connection cancel];
+            NSError* failingError = [NSError errorWithDomain:NSURLErrorDomain
+                                                        code:-1003
+                                                    userInfo:@{ NSURLErrorFailingURLErrorKey : request.URL }];
+            [self.client URLProtocol:self didFailWithError:failingError];
```
jasoncoolmax: do we need the code to handle the end url case in
- (void)webAuthDidFailWithError:(NSError *)error?

RPangrle: I don't think so, we had to do that before because we weren't getting hit with shouldStartLoadRequest: I'll know for sure when I get a test pass done.

jasoncoolmax: Oh, you are right :)

jasoncoolmax:
:shipit:

PR #764, opened 2016-09-11
Add check for end URL in the webview delegate

RPangrle:
We do still hit this when the redirect URI is an "http://" URL

### DIFF
--- a/ADALiOS/ADALiOS/ADAuthenticationWebViewController.m
+++ b/ADALiOS/ADALiOS/ADAuthenticationWebViewController.m
@@ -215,6 +215,19 @@ NSTimer *timer;
         return;
     }
     
+    // If we failed on an invalid URL check to see if it matches our end URL
+    if ([error.domain isEqualToString:@"NSURLErrorDomain"] && (error.code == -1002 || error.code == -1003))
+    {
+        NSURL* url = [error.userInfo objectForKey:NSURLErrorFailingURLErrorKey];
+        NSString* urlString = [url absoluteString];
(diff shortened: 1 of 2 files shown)


ADALiOS/ADALiOS/ADAuthenticationWebViewController.m
```diff
@@ -215,6 +215,19 @@ - (void)webView:(UIWebView *)webView didFailLoadWithError:(NSError *)error
         return;
     }
     
+    // If we failed on an invalid URL check to see if it matches our end URL
+    if ([error.domain isEqualToString:@"NSURLErrorDomain"] && (error.code == -1002 || error.code == -1003))
```
RandalliLama: @RPangrle  Can you define constants for these error codes that have a descriptive name?

RandalliLama:
@RPangrle  Is there an issue on this?  Can you describe in more detail why this is necessary?

jasoncoolmax:
I signed off :)

RPangrle:
This was issue #720 in 2.x

RandalliLama:
I'm not clear on what the description in #720 means.  Can you guys elaborate?  In issue #720?

RPangrle:
We fail any attempt to navigate to a non-secure URL, however if that URL was the redirect URI we should still succeed. This extra bit of code is checking in the error handling to code to see if the failing URL is our redirect URI.

RandalliLama:
@RPangrle Got it.  I've update #720 with that information.

RandalliLama:
:shipit:

PR #769, opened 2016-09-13
convergence iOS 10 Support

RPangrle:
Remove NTLM
Fix a percent-unencoding bug in -[ADHelpers addClientVersionToURL:]


jasoncoolmax:
:shipit:

PR #775, opened 2016-09-15
add quotes to the run script arguments

### LINKED ISSUE #774: Run Script in build phase not run property when project dir contains spaces
This is because there is no quotes on the run script arguments.

jasoncoolmax:
fixes #774 

Now the quotes have been added to the run script arguments. Otherwise it won't run properly when project dir path contains spaces.


RPangrle:
Welcome to the fun of bash scripting :)

:shipit:

PR #777, opened 2016-09-16
Fix private Apple api name collision

eddiekim:
Fix app submission review because of name collision with private API `dispatch:`


jasoncoolmax:
I signed off :)

Issue #773, opened 2016-09-14
Crash in internalHandleBrokerResponse: when we try to process something that is not actually a broker response


joeydong:
The Outlook app switches to the Authenticator app to collect Intune diagnostic logs. The last step of that process is to app switch from Authenticator back to Outlook.

The problem is that ADAL is catching this `application:openURL` in Outlook and attempting to handle a call from Authenticator when auth is not currently in progress. This leads to a crash because ADAL attempts to execute a non-existent `completionBlock`.

Here's an abbreviated crash log
```

Thread 0 Crashed:
0   ADAL                                 0x00000001019d60d8 +[ADAuthenticationContext(Internal) checkAndHandleBadArgument:argumentName:correlationId:completionBlock:] (ADAuthenticationContext+Internal.m:63)
1   ADAL                                 0x00000001019b67c8 +[ADAuthenticationRequest(Broker) internalHandleBrokerResponse:] (ADAuthenticationRequest+Broker.m:96)
2   ADAL                                 0x00000001019d89f4 __swizzle_ApplicationOpenURL (ADBrokerHelper.m:51)
3   UIKit                                0x00000001919a0924 __45-[UIApplication _applicationOpenURL:payload:]_block_invoke + 868
4   UIKit                                0x00000001919a0338 -[UIApplication _applicationOpenURL:payload:] + 652
5   UIKit                                0x00000001919a8588 -[UIApplication _handleNonLaunchSpecificActions:forScene:withTransitionContext:completion:] + 4304
6   UIKit                                0x00000001919acfcc -[UIApplication _handleApplicationActivationWithScene:transitionContext:completion:] + 2408
7   UIKit                                0x00000001919ac3ec -[UIApplication _handleApplicationLifecycleEventWithScene:transitionContext:completion:] + 448
8   UIKit                                0x0000000191997ae8 __70-[UIApplication scene:didUpdateWithDiff:transitionContext:completion:]_block_invoke + 148
9   UIKit                                0x000000019199776c -[UIApplication scene:didUpdateWithDiff:transitionContext:completion:] + 884
10  UIKit                                0x0000000191cc9034 -[UIApplicationSceneClientAgent scene:handleEvent:withCompletion:] + 460
11  FrontBoardServices                   0x000000018d43bbd4 __80-[FBSSceneImpl updater:didUpdateSettings:withDiff:transitionContext:completion:]_block_invoke.376 + 204
[...]

PR #779, opened 2016-09-19
Add more robust parsing of broker response messages

RPangrle:
- Remove the usage of completionBlock while parsing the broker's response. This lead to some incorrect usage of error checking macros that assumed that completionBlock could never be nil.
- Retain state from the initial broker request that can persist across app launches so that we can use that in the validation code, and in the future better support for caching results when we receive a broker response on a fresh app launch (the app was killed during the broker request)
- If the URL is not a proper broker response then fall through to the app's original URL handler to allow them to process it.
- Add test coverage of handleBrokerResponse to cover these scenarios and a valid broker response
- Modifications to ADBrokerKeyHelper to wrangle the key used for crypto in the tests.

### DIFF
--- a/ADAL/src/request/ADAuthenticationRequest+Broker.m
+++ b/ADAL/src/request/ADAuthenticationRequest+Broker.m
@@ -327,13 +327,13 @@ NSString* kAdalResumeDictionaryKey = @"adal-broker-resume-dictionary";
       @"extra_qp"       : _queryParams ? _queryParams : @"",
       };
     
-    NSDictionary* resumeDictionary =
+    NSDictionary<NSString *, NSString *>* resumeDictionary =
   @{
     @"authority"        : _context.authority,
     @"resource"         : _resource,
     @"client_id"        : _clientId,
     @"redirect_uri"     : _redirectUri,
-    @"correlation_id"   : _correlationId,
+    @"correlation_id"   : _correlationId.UUIDString,
     };
     
     [[NSUserDefaults standardUserDefaults] setObject:resumeDictionary forKey:kAdalResumeDictionaryKey];
--- a/ADAL/src/request/ADAuthenticationRequest+Broker.m
+++ b/ADAL/src/request/ADAuthenticationRequest+Broker.m
@@ -85,8 +85,8 @@ NSString* kAdalResumeDictionaryKey = @"adal-broker-resume-dictionary";
     ADAuthenticationResult* result = [self processBrokerResponse:response];
     
     [[NSUserDefaults standardUserDefaults] removeObjectForKey:kAdalResumeDictionaryKey];
-    // If we didn't a authentication result then assume it wasn't actually a broker message. This
-    // allows the message to pass through to other handlers
+    // If we didn't get an authentication result then assume it wasn't actually a broker
+    // message. This allows the message to pass through to other handlers
     if (!result)
     {
         if (completionBlock)
--- a/ADAL/src/request/ADAuthenticationRequest+Broker.m
+++ b/ADAL/src/request/ADAuthenticationRequest+Broker.m
@@ -42,6 +42,8 @@
 #import "ADBrokerNotificationManager.h"
 #endif // TARGET_OS_IPHONE
 
+NSString* kAdalResumeDictionaryKey = @"adal-broker-resume-dictionary";
+
 @implementation ADAuthenticationRequest (Broker)
 
 + (BOOL)validBrokerRedirectUri:(NSString*)url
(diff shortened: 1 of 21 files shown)


ADAL/src/request/ADAuthenticationRequest+Broker.m
```diff
@@ -75,11 +77,78 @@ + (BOOL)validBrokerRedirectUri:(NSString*)url
...
 {
 #if TARGET_OS_IPHONE
-    ADAuthenticationCallback completionBlock = [ADBrokerHelper copyAndClearCompletionBlock];
-    HANDLE_ARGUMENT(response, nil);
+    __block ADAuthenticationCallback completionBlock = [ADBrokerHelper copyAndClearCompletionBlock];
+        
+    ADAuthenticationResult* result = [self processBrokerResponse:response];
+    
+    [[NSUserDefaults standardUserDefaults] removeObjectForKey:kAdalResumeDictionaryKey];
+    // If we didn't a authentication result then assume it wasn't actually a broker message. This
```
jasoncoolmax: If we didn't **get** a ....?

ADAL/src/request/ADAuthenticationRequest+Broker.m
```diff
@@ -238,22 +309,34 @@ - (void)callBroker:(ADAuthenticationCallback)completionBlock
...
+    NSDictionary* resumeDictionary =
+  @{
+    @"authority"        : _context.authority,
+    @"resource"         : _resource,
+    @"client_id"        : _clientId,
+    @"redirect_uri"     : _redirectUri,
+    @"correlation_id"   : _correlationId,
+    };
+    
+    [[NSUserDefaults standardUserDefaults] setObject:resumeDictionary forKey:kAdalResumeDictionaryKey];
```
jasoncoolmax: do we need to set this flag in function handleBrokerFromWebiewResponse as well?

RPangrle: @kpanwar Do we know what scenario we would even hit that codepath? It looks like it happens when we go through an auth code flow and instead get hit with "msauth://"--- the only way I could even think of hitting that is if the application isn't configured to use broker--- and yet the server decides to try and force it anyways? This doesn't seem like something we should be supporting.

ADAL/src/request/ADAuthenticationRequest+Broker.m
```diff
@@ -75,11 +77,78 @@ + (BOOL)validBrokerRedirectUri:(NSString*)url
...
++ (BOOL)internalHandleBrokerResponse:(NSURL *)response
 {
 #if TARGET_OS_IPHONE
-    ADAuthenticationCallback completionBlock = [ADBrokerHelper copyAndClearCompletionBlock];
-    HANDLE_ARGUMENT(response, nil);
+    __block ADAuthenticationCallback completionBlock = [ADBrokerHelper copyAndClearCompletionBlock];
+        
+    ADAuthenticationResult* result = [self processBrokerResponse:response];
+    
+    [[NSUserDefaults standardUserDefaults] removeObjectForKey:kAdalResumeDictionaryKey];
```
jasoncoolmax: do we need to remove the flag in function callbackCleanup of ADBrokerNotificationManager.m too?

ADAL/src/request/ADAuthenticationRequest+Broker.m
RPangrle: The two paths are nearly identical... except one never got updated to support version 2, and will prompt install, so it was pretty easy to collapse the two into a single callBroker:

ADAL/src/request/ADAuthenticationRequest+Broker.m
RPangrle: Nope. openURL will get hit first if we're hit by a URL. In other cases, if another broker request happens it'll overwrite the dictionary, and if we end up getting hit with a broker response afterwards somehow the data could be useful for dropping it in the proper cache location. (If we choose to do such a thing) So in either case leaving around the data isn't harmful and possibly useful.

jasoncoolmax: Make sense :)

jasoncoolmax:
:shipit:

ADAL/src/request/ADAuthenticationRequest+Broker.m
RPangrle: As a general rule I prefer to avoid doing extra cleaning-of-state unless we can think of a case where it'd be necessary.

RPangrle:
Fixed #773

Issue #778, opened 2016-09-19
NSNotificationCenter observer memory leak


Neil3079:
Whilst profiling my app it seems like near the [ADBrokerHelper load] method some notification observers are not getting deallocated correctly causing leaks

<img width="941" alt="adalleakone" src="https://cloud.githubusercontent.com/assets/2815318/18628730/82c80cde-7e59-11e6-9122-bf3fc01b24e4.png">

There seem to be a few more of these leaks when invoking the framework.

I am using the ADAL framework version 2.2.5 in an Xcode 8, Swift 3 project. The framework is installed using cocoapods

PR #780, opened 2016-09-19
Fix memory leak of observer by declaring __weak

RPangrle:
Otherwise ARC holds a strong reference to observer which creates a retain cycle


jasoncoolmax:
:shipit:

RPangrle:
Fixed #778

Issue #695, opened 2016-07-11
Add Telemetry feature


jasoncoolmax:
High level goal is to enable reporting on AuthN experience, key scenarios 
1) too many login prompts, 
2) login failure,
3) slow performance.

eddiekim:
Would a client be able to measure token refresh timing with this? Specifically, we need elapsed time for a network token refresh. If not, I can open a new issue to discuss.

eddiekim:
> Would a client be able to measure token refresh timing with this? Specifically, we need elapsed time for a network token refresh. If not, I can open a new issue to discuss.

Should I create a new issue to discuss this?

jasoncoolmax:
Sorry that I didn't notice that there is a question here :package: 

Yes, Telemetry does generate event for token refresh(with information like start time, end time, etc.)

PR #782, opened 2016-09-20
Apply telemetry interface to existing code

jasoncoolmax:
(for #695)

Telemetry events are generated by applying the telemetry hooks to existing code.

The following 5 types of events are generated:

1) ADAPIEvent (for api calls to reflect the work flow)
2) ADUIEvent (for webview launch)
3) ADHttpEvent (for http connection)
4) ADCacheEvent (for cache access)
5) ADBrokerEvent (for broker invoke)

Besides, 
-  Class ADRequestParameters is added to wrap up auth request parameters, like authority, resource, client id etc.
- Changes has been made to generate device id for Mac OS, which is a hash value of the Mac serial number.


RPangrle:
:shipit:

PR #783, opened 2016-09-21
Object Model changes for broker refactor

RPangrle:
Typedef the completionBlock type for ADWebAuthRequest
Define an ADRequestContext protocol for passing through the correlation ID and telemetry ID into internal ADAL APIs
Change many internal ADAL APIs to accept a id<ADRequestContext>
Fix ADUserIdentifier to conform to NSCopying


jasoncoolmax:
:shipit:

PR #793, opened 2016-09-29
Forward `application:openURL:` call

joeydong:
Following up with https://github.com/AzureAD/azure-activedirectory-library-for-objc/pull/779 and https://github.com/AzureAD/azure-activedirectory-library-for-objc/issues/773:

Improve `ADBrokerHelper` to call original `application:openURL:` if ADAL fails to handle it. In some cases, calls to `application:openURL` from the broker app are not intended for ADAL.

This way the original host application can access the URL parameters of the `application:openURL:` call and respond appropriately.


RPangrle:
Can you please target this fix into the release/v2.2.7 branch instead?

joeydong:
Yup, sorry for the delay. What do you think about the changes?

RPangrle:
Woah. You can change the base branch now without having to reopen the PR? The changes look good. :shipit:

PR #802, opened 2016-10-07
Fine Grained Broker Processing Errors

RPangrle:
More fine grained error reporting and clearer error pathways for the broker response code
-[NSUserDefaults synchronize] call after storing resume dictionary


jasoncoolmax:
:shipit:

PR #803, opened 2016-10-10
Don't cause errors on Mac when get WPJ registration

RPangrle:
Some ADFS servers can attempt pkeyauth all the time. Always failling on any pkeyauth attempt on Mac causes those auth flows to fail even when they still might be able to pass through other means.


RPangrle:
Did the polar opposite of what I was trying to do in this PR...

PR #804, opened 2016-10-10
Don't create a WPJ-not-supported error on Mac

RPangrle:
This causes the authentication flow to error out early whenever we get a pkeyauth challenge on Mac. Some ADFS servers are configured to always send pkeyauth challenges even if they aren't mandatory. Not erroring out here lets us finish the authentication fflow through other methods.


jasoncoolmax:
:shipit:

PR #808, opened 2016-10-10
Add +[ADTokenCache defaultTokenCache] convenience method

RPangrle:
This method returns a singleton for the default token cache using the default storage delegate. This should make it a little easier for application developers to discover how to manipulate the cache on MacOS.

### DIFF
--- a/ADAL/src/public/mac/ADTokenCache.h
+++ b/ADAL/src/public/mac/ADTokenCache.h
@@ -46,6 +46,10 @@
     pthread_rwlock_t _lock;
 }
 
+/*! Returns the default cache object using the ADTokenCacheDelegate set in
+    ADAuthenticationSettings */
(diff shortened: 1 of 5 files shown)


ADAL/src/public/mac/ADTokenCache.h
```diff
@@ -46,6 +46,10 @@
     pthread_rwlock_t _lock;
 }
 
+/*! Returns the default cache object using the ADTokenCacheDelegate set in
+    ADAuthenticationSettings */
```
jasoncoolmax: did we forget to set the delegate in function _defaultCache_?

RPangrle: -[ADAuthenticationSettings setDefaultStorageDelegate:] will set it on +[ADTokenCache defaultCache]

jasoncoolmax:
:shipit:

PR #811, opened 2016-10-18
Strip off the keychain team ID from the keychain sharing group before…

### LINKED ISSUE #810: Keychain "add" operation fails from brokered auth
Users are unable to stay logged in because the brokered auth flow fails to save tokens to keychain. Non-broker flow is fine.

branch: `release/2.2.7`

App logs:

> ADAL 2.2.7 iOS 10.0.2 [2016-10-14 20:01:12] Error raised: (Domain: "ADKeychainErrorDomain" Code: -34018 ProtocolCode: "(null)" Details: "Keychain failed during "add" operation" (errorCode=-34018, additionalInfo=(null), userInfo={
    error = "Error with code: -34018 Domain: ADKeychainErrorDomain ProtocolCode:(null) Details:Keychain failed during \"add\" operation. Inner error details: Error Domain=ADKeychainErrorDomain Code=-34018 \"(null)\"";
})

Device console: 

> add Error Domain=NSOSStatusErrorDomain Code=-34018 "explicit accessGroup TEAMID.TEAMID.com.microsoft.adalcache not in client access

RPangrle:
… saving it in the resume dictionary

Fixes #810

### DIFF
--- a/ADAL/src/request/ADAuthenticationRequest+Broker.m
+++ b/ADAL/src/request/ADAuthenticationRequest+Broker.m
@@ -40,6 +40,7 @@
 #import "ADKeychainTokenCache+Internal.h"
 #import "ADBrokerKeyHelper.h"
 #import "ADBrokerNotificationManager.h"
+#import "ADKeychainUtil.h"
 #endif // TARGET_OS_IPHONE
 
 NSString* kAdalResumeDictionaryKey = @"adal-broker-resume-dictionary";
@@ -306,6 +307,17 @@ NSString* kAdalResumeDictionaryKey = @"adal-broker-resume-dictionary";
     id<ADTokenCacheDataSource> dataSource = [_tokenCache dataSource];
     if (dataSource && [dataSource isKindOfClass:[ADKeychainTokenCache class]])
     {
+        NSString* keychainGroup = [(ADKeychainTokenCache*)dataSource sharedGroup];
+        NSString* teamId = [ADKeychainUtil keychainTeamId:&error];
+        if (!teamId && error)
+        {
+            completionBlock([ADAuthenticationResult resultFromError:error]);
+            return;
+        }
+        if (teamId && [keychainGroup hasPrefix:teamId])
+        {
+            keychainGroup = [keychainGroup substringFromIndex:teamId.length + 1];
+        }
         resumeDictionary =
         @{
           @"authority"        : _context.authority,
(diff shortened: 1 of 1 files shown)


RPangrle:
@joeydong

ADAL/src/request/ADAuthenticationRequest+Broker.m
```diff
@@ -306,14 +307,25 @@ - (void)callBroker:(ADAuthenticationCallback)completionBlock
     id<ADTokenCacheDataSource> dataSource = [_tokenCache dataSource];
     if (dataSource && [dataSource isKindOfClass:[ADKeychainTokenCache class]])
     {
+        NSString* keychainGroup = [(ADKeychainTokenCache*)dataSource sharedGroup];
+        NSString* teamId = [ADKeychainUtil keychainTeamId:&error];
```
joeydong: is it possible for `[ADKeychainUtil keychainTeamId:&error];` to return something like an empty string and also an error?

joeydong: I traced through the code a little bit and I guess we're fine because you check the result length in `ADKeychainUtil.retrieveTeamIDFromKeychain`.

This leads me to another question: The query you use in `ADKeychainUtil.retrieveTeamIDFromKeychain` searches for a keychain item with `teamIDHint` but `ADKeychainTokenCache` inserts a keychain item with no label. Are these two separate systems?

joeydong:
I left some comments directly on https://github.com/AzureAD/azure-activedirectory-library-for-objc/commit/8ddc3ae8a47ac01c6b8e5cfd07e0da6f39e73bc3 (in case you didn't get notified about it)

ADAL/src/request/ADAuthenticationRequest+Broker.m
RPangrle: +[ADKeychainUtil retrieveTeamIDFromKeychain] will create the entry it is looking for if it doesn't exist.

And, no, keychainTeamId will not return an empty string and an error either. The error it returns is from retrieveTeamIDFromKeychain when they keychain operation fails. In that case the keychainTeamID never gets set.

jasoncoolmax:
:shipit:

ADAL/src/request/ADAuthenticationRequest+Broker.m
joeydong: Okay sounds good. I forgot the method was called `retrieveTeamIDFromKeychain` so it doesn't matter if the rest of the app group name isn't the right "adalcache" one.

iamshezad:
"ADKeychainErrorDomain" Code: -34018 ProtocolCode: "(null)" Details: "Keychain failed during "add" operation"
Error: %@ Could not connect to the server.
How to solve this errors???

harkmall:
+1 still getting this error on iOS 11

PR #815, opened 2016-10-21
Keychain and Result logging improvements

RPangrle:
I'm trying to clear up confusion over specific log line (i.e. "Nothing found in keychain" which simply meant none were found for the current query, not that they all disappeared...) Increased noise from the FoCI error fallback codepaths have made it more difficult to tell what the final result of an acquireToken call is, so I have added a definitive end log statement for acquireToken flows, as multiple errors can occur in a acquireToken flow, but it still can succeed.

* Log clear acquireToken end messages so we have definitive ADAL logs for the eventual error that was sent to the developer
* Move context information for keychain queries into the "info" section of the keychain logging

### DIFF
--- a/ADAL/src/cache/ios/ADKeychainTokenCache.m
+++ b/ADAL/src/cache/ios/ADKeychainTokenCache.m
@@ -193,20 +193,16 @@ static ADKeychainTokenCache* s_defaultCache = nil;
                         userId:(NSString *)userId
                  correlationId:(NSUUID *)correlationId
 {
+    NSString* keyCtxStr = [NSString stringWithFormat:@"(resource <%@> + client <%@> + authority <%@>)", [key resource], [key clientId], [key authority]];
     if (!items || [items count]<=0)
     {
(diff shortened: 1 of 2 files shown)


ADAL/src/cache/ios/ADKeychainTokenCache.m
```diff
@@ -241,9 +237,13 @@ - (NSString*)getTokenNameForLog:(ADTokenCacheItem *)item
             [tokenName stringByAppendingString:@"+RT"];
         }
     }
+    else if ([item.clientId hasPrefix:@"foci-"])
+    {
+        tokenName = @"FRT";
+    }
     else if (![NSString adIsStringNilOrBlank:item.refreshToken] && [NSString adIsStringNilOrBlank:item.resource])
     {
-        tokenName = @"MRRF";
```
jasoncoolmax: My bad, silly typo T_T

jasoncoolmax:
:shipit:

PR #816, opened 2016-10-21
Improve token logging

RPangrle:
A number of the token log statements don't provide useful context for what that line is for, and often are being logged at verbose, instead of info. Also, shorten the hashes we use to log. 40 character hashes aren't needed for logging purposes, the first 7 characters will be more then sufficient.


jasoncoolmax:
:shipit:

PR #822, opened 2016-11-03
Error out when extraQueryParameters are not in good form

### LINKED ISSUE #819: Crash with Chinese Characters
`ADAuthenticationContext.acquireTokenWithResource` includes an `extraQueryParameters` parameter. We are seeing an exception / crash when Chinese characters are used in the domain name of the `login_hint` passed in the `extraQueryParameters` parameter.

The result is that the `startUrl` is nil when `ADWebAuthController` is initialized and an exception is thrown, crashing the app.

Instead of throwing an exception (in non DEBUG environments), this should make it's way back up the stack and be returned as `success = false` in the `ADAuthenticationContext.acquireTokenWithResource` call.

jasoncoolmax:
(fixes #819)

If Webview startURL fails to construct because of extraQueryParameters, AD_ERROR_DEVELOPER_INVALID_ARGUMENT will be returned.

Unit test is also added for testing bad extraQueryParameters.

### DIFF
--- a/ADAL/src/request/ADAuthenticationRequest+AcquireToken.m
+++ b/ADAL/src/request/ADAuthenticationRequest+AcquireToken.m
@@ -77,6 +77,17 @@
         return;
     }
     
+    if (![self checkExtraQueryParameters])
+    {
+        ADAuthenticationError* error =
+        [ADAuthenticationError errorFromAuthenticationError:AD_ERROR_DEVELOPER_INVALID_ARGUMENT
+                                               protocolCode:nil
+                                               errorDetails:@"extraQueryParameters is not properly encoded. Please make sure it is URL encoded."
+                                              correlationId:_correlationId];
+        wrappedCallback([ADAuthenticationResult resultFromError:error correlationId:_correlationId]);
+        return;
+    }
+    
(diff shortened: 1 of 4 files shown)


RPangrle:
These changes are a good start, I would encourage you to add a check on the redirect URI when creating the ADAuthenticationRequest and send a specific error letting the developer know that's why the request failed.

jasoncoolmax:
Sure :)

RPangrle:
Sorry, I meant the EQPs, not redirect URI, sorry about that.

RPangrle:
:shipit:

ADAL/src/request/ADAuthenticationRequest+AcquireToken.m
```diff
@@ -111,6 +122,23 @@ - (void)acquireToken:(ADAuthenticationCallback)completionBlock
...
 
+- (BOOL)checkExtraQueryParameters
+{
+    if ([NSString adIsStringNilOrBlank:_queryParams])
+    {
+        return YES;
+    }
+    
+    NSString* queryParams = _queryParams.adTrimmedString;
+    if ([queryParams hasPrefix:@"&"])
```
jasoncoolmax: not sure, but in case somebody will do it that way

PR #823, opened 2016-11-05
"Get the App" crash fix

### LINKED ISSUE #818: Crash from pressing "Get the App" in NSURL
There is a nil string passed as a parameter into the NSURL initializer.

It’s a 100% repro when the “Get the App” button is pressed using the following credentials:

<redacted>

Stack trace and exception are attached.

adal-crash-stack

RPangrle:
Rename -callBroker into -composeBrokerRequest so it can be better focused on a single scenario, and then call invokeBroker or promptBrokerInstall as appropriate.

Fixes #818


jasoncoolmax:
:shipit:

PR #829, opened 2016-11-10
nil out _authenticationViewController when web view timeout happens

### LINKED ISSUE #828: _authenticationViewController does not nil out when webview timeout happens
This will affect the authenticator SDK in the following way:

If `_authenticationViewController` does not nil out when webview timeout happens,
when the next time `invokeBroker` is being called, it will find that there is an ongoing web auth session going on (which is not true) and will wait for it to end forever.

jasoncoolmax:
fixes #828


RPangrle:
Any idea of this is needed in 2.3/dev?

RPangrle:
:shipit:

PR #833, opened 2016-11-14
bug fix for broker sdk refactoring

jasoncoolmax:
- copy over the fixes from #828 to dev
- fix a bug where completionBlock is not executed when `returnRawResponse=YES`

### DIFF
--- a/ADAL/src/request/ADWebAuthResponse.m
+++ b/ADAL/src/request/ADWebAuthResponse.m
@@ -106,6 +106,7 @@
                 [_responseDictionary setObject:rawResponse
                                         forKey:@"raw_response"];
                 SAFE_ARC_RELEASE(rawResponse);
+                completionBlock(_responseDictionary);
                 break;
             }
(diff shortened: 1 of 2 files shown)


ADAL/src/request/ADWebAuthResponse.m
```diff
@@ -106,6 +106,7 @@ - (void)handleResponse:(ADWebResponse *)webResponse
                 [_responseDictionary setObject:rawResponse
                                         forKey:@"raw_response"];
                 SAFE_ARC_RELEASE(rawResponse);
+                completionBlock(_responseDictionary);
                 break;
```
RPangrle: Make this a return; not a break; then. Otherwise someone might change the subsequent code later and cause other regressions.

jasoncoolmax: Sure :)

RPangrle:
:shipit:

PR #838, opened 2016-11-17
Update for Carthage compatibility

RPangrle:
Including ADAL as a subproject in the the SampleSwiftApp project confused the carthage tool, combined with xcodebuild bugs it would block Carthage from working.


jasoncoolmax:
:shipit:

PR #839, opened 2016-11-17
disable timeout for initial NSURLRequest before popping up web view

jasoncoolmax:


ADALiOS/ADALiOS/ADWebAuthController.m
```diff
@@ -605,7 +605,9 @@ - (void)start:(NSURL *)startURL
         _completionBlock(error, nil);
     }
     
-    NSMutableURLRequest* request = [[NSMutableURLRequest alloc] initWithURL:[ADHelpers addClientVersionToURL:startURL]];
+    NSMutableURLRequest* request = [[NSMutableURLRequest alloc] initWithURL:[ADHelpers addClientVersionToURL:startURL]
+                                                                cachePolicy:NSURLRequestUseProtocolCachePolicy
+                                                            timeoutInterval:0];
```
RPangrle: These two changes seem contradictory to me. Either all of them should timeout at 60s or all should have timeouts disabled...

ADALiOS/ADALiOS/ADURLProtocol.m
```diff
@@ -137,6 +137,7 @@ - (void)startLoading
     AD_LOG_VERBOSE_F(@"-[ADURLProtocol startLoading]", @"host: %@", [self.request.URL host]);
     
     NSMutableURLRequest* request = [self.request mutableCopy];
+    [request setTimeoutInterval:60.0];//set default timeout interval after copy
```
jasoncoolmax: According to my test, the above line makes no difference to the behaviour. 
In ADURLProtocol, no timeout will be triggered with or without the above line, or even set `[request setTimeoutInterval:10.0];`.

So I tried to keep the change "as minimal as possible", making sure I won't change any behavior of ADURLProtocol.

ADALiOS/ADALiOS/ADWebAuthController.m
jasoncoolmax: You are right. I have left some comments below :)

ADALiOS/ADALiOS/ADURLProtocol.m
jasoncoolmax: I have some updates on this part :)

kpanwar:
🚢 it.

PR #842, opened 2016-11-24
Releases parentViewController in ADAuthenticationBroker

alexanderlarionov:
Hello! Faced the problem when authViewController's presentingController (the top one in UI hierarchy of my app) is not being dealloced. It happens because ADAuthenticationBroker retains reference to it, but never releases. So I fixed it, not it releases parentViewController in ADAuthenticationBroker on webView success/fail/cancel.

Thank you!


RPangrle:
This would have to go into a 1.2.x-dev branch, however 1.2.x is in maintenance hotfix-only mode now, and we have no plans for another 1.2.x release. We encourage you to move to the latest 2.x version.

PR #844, opened 2016-11-30
Print hashes in the token item description

RPangrle:
This causes hashes to show up in our results box which makes it a lot easier to quickly verify that a new token indeed came back.


jasoncoolmax:
:shipit:

PR #845, opened 2016-11-30
Mac test app improvements

RPangrle:
Expanded out the cache window
Set better truncation settings on cache columns
Added clear cache and clear cookies buttons to the acquire token window


jasoncoolmax:
:shipit:

PR #848, opened 2016-12-02
Add "--no-xcpretty" switch

RPangrle:
Sometimes xcpretty doesn't print out output, which can make things difficult to debug.


jasoncoolmax:
:shipit:

PR #849, opened 2016-12-03
Dispatch to the main thread first and create the alert

RPangrle:
The associated object call created some memory management issues when the server is misbehaving.


ADAL/src/ui/ios/ADNTLMUIPrompt.m
```diff
@@ -21,26 +21,60 @@
...
+        UITextField* username = alert.textFields.firstObject;
+        UITextField* password = alert.textFields.lastObject;
+        
+        block(username.text, password.text);
     }];
+    
+    [alert addAction:cancelAction];
+    [alert addAction:loginAction];
+    
+    [alert addTextFieldWithConfigurationHandler:^(UITextField * _Nonnull textField) { (void)textField; }];
```
jasoncoolmax: Shall we add a placeholder like "Username" and "Password"? Currently it seems empty for both textboxes :)

RPangrle: MacOS UI has labels on it, placeholders are less common.

jasoncoolmax: OK :)

jasoncoolmax:
:shipit:

ADAL/src/ui/ios/ADNTLMUIPrompt.m
RPangrle: Oh wait, sorry, this is iOS UI. The Safari version of this control doesn't user placeholders and then we would have to localize those strings, so it's not worth it.

jasoncoolmax:
:shipit:

PR #854, opened 2016-12-12
Mac and iOS Test App Consistency Changes

RPangrle:
- Name buttons "acquireToken" and "acquireTokenSilent" on both platforms
- Added a profile chooser button on the acquire page of the iOS test app
- Consistent result strings for clearing the cache on iOS and Mac
- Removed IBOutlets from code no longer using XIBs
- Fixed a minor bug in scrollview caused by not using autolayout through the entire view heirarchy.


jasoncoolmax:
:shipit:

PR #859, opened 2016-12-16
Improve Telemetry

jasoncoolmax:
- Improvement on Telemetry public header.

- Due to some design change, event is now represented by a dictionary.

- Fix a bug where broker app/protocol version is not properly parsed, a place where previous http event properties could be left during aggregation.

- Fix a bug for setting auth error code

- Strengthen unit tests


ADAL/src/telemetry/ADTelemetryAPIEvent.m
```diff
@@ -162,29 +189,50 @@ - (void)addAggregatedPropertiesToDictionary:(NSMutableDictionary *)eventToBeDisp
...
         {
-            [eventToBeDispatched setObject:property.value forKey:property.name];
+            [eventToBeDispatched setObject:[properties objectForKey:name] forKey:name];
         }
     }
 }
 
+- (void)initAPIEventProperties
+{
+    [self setProperty:AD_TELEMETRY_RESULT_STATUS value:@""];
```
kpanwar: why are these properties empty? A property should either exist with the expected value or it should not. this applies to all init calls in events. @addev-ashish - do you agree with this comment?

jasoncoolmax: It is said that each event not having all properties is a bug, such that one event cannot fully overwrite another event during integration. 

If we don't want it that way, I also have other solutions to make sure one event can fully overwrite another. But I am curious how other platforms solve this issue? :)

ADAL/src/request/ADAuthenticationRequest+Broker.m
```diff
@@ -205,6 +203,7 @@ + (ADAuthenticationResult *)processBrokerResponse:(NSURL *)response
     {
         protocolVersion = [msgVer integerValue];
     }
+    s_brokerProtocolVersion = msgVer;
```
RPangrle: Having this as a static variable makes me nervous. It really shouldn't be.

ADAL/src/telemetry/ADTelemetryDefaultEvent.m
```diff
@@ -135,22 +134,19 @@ - (NSString*)getStringFromDate:(NSDate*)date
...
 - (void)addAggregatedPropertiesToDictionary:(NSMutableDictionary*)eventToBeDispatched
 {
-    for (int i=0; i<[self getDefaultPropertyCount]; i++)
-    {
-        NSString* propertyName = [(ADTelemetryProperty*)_propertyMap[i] name];
-        NSString* propertyValue = [(ADTelemetryProperty*)_propertyMap[i] value];
-        [eventToBeDispatched setObject:propertyValue forKey:propertyName];
-    }
+    [eventToBeDispatched addEntriesFromDictionary:[self defaultParameters]];
+    SET_IF_NOT_NIL(eventToBeDispatched, AD_TELEMETRY_REQUEST_ID, [_propertyMap objectForKey:AD_TELEMETRY_REQUEST_ID]);
```
RPangrle: Why are these values being aggregated? Aggregation by definition can be across multiple requests and correlation IDs.

ADAL/src/telemetry/ADTelemetryEventInterface.h
```diff
@@ -23,11 +23,11 @@
 
 @protocol ADTelemetryEventInterface <NSObject>
 
-@property (readonly) NSMutableArray* propertyMap;
+@property (readonly) NSMutableDictionary* propertyMap;
```
RPangrle: Readonly mutable dictionary is something of an oxymoron...

ADAL/src/telemetry/ADTelemetryAPIEvent.m
```diff
@@ -162,27 +162,26 @@ - (void)addAggregatedPropertiesToDictionary:(NSMutableDictionary *)eventToBeDisp
 {
     [super addAggregatedPropertiesToDictionary:eventToBeDispatched];
     
-    NSArray* properties = [self getProperties];
-    for (ADTelemetryProperty* property in properties)
+    NSDictionary* properties = [self getProperties];
+    for (NSString* name in properties)
```
RPangrle: Also, this logic easily could be genericized to take in an array of properties as well....

jasoncoolmax: 👍 Nice! let me make the change!

ADAL/src/telemetry/ADTelemetryDefaultEvent.m
jasoncoolmax: currently the aggregation feature by design is to aggregate all relevant events related to a particular acquireToken call into one single event.

ADAL/src/request/ADAuthenticationRequest+Broker.m
jasoncoolmax: I am also thinking the same :) 
But I am having trouble to inject the broker telemetry event there. When broker returns back, ADBrokerHelper calls the class method of ADAuthenticationRequest to handle the response from broker.

ADAL/src/public/ADTelemetry.h
```diff
@@ -76,15 +57,15 @@
...
++ (nonnull ADTelemetry*)sharedInstance;
 
 /*!
     Register a telemetry dispatcher for receiving telemetry events.
     @param dispatcher            An instance of ADDispatcher implementation.
     @param aggregationRequired   Specifies if telemetry events will be aggregated on client, 
                                  i.e. all events of a single request will be aggregated as one single event.
  */
-- (void)registerDispatcher:(id<ADDispatcher>)dispatcher
+- (void)registerDispatcher:(nullable id<ADDispatcher>)dispatcher
```
eddiekim: shouldn't this be non null?

eddiekim:
I have a small request:

Can you add the event names to `ADTelemetryEventStrings` ? In our case, we're interested in a particular event (`http_request`), and it'd be nice to use a defined variable rather than matching the hardcoded event name that ADAL uses internally.

ADAL/src/public/ADTelemetry.h
```diff
@@ -34,28 +34,9 @@
...
-{
-    NSString* _name;
-    NSString* _value;
-}
-
-@property (readonly) NSString* name;
-@property (readonly) NSString* value;
-
-- (id)initWithName:(NSString*)name value:(NSString*)value;
+- (void)dispatchEvent:(nonnull NSDictionary<NSString*, NSString*> *)event;
```
eddiekim: Naming suggestion: aren't these actually properties of the event? It'd make the argument a little clearer to call it `properties` rather than `event`.

alternatively, why not just pass the `ADTelemetryEventInterface` object to the delegate? The interface is a little confusing because you expect to receive an "event", but instead you get a bunch of properties that describe the event.

ADAL/src/public/ADTelemetry.h
RPangrle: Dispatchers can be unregistered.

ADAL/src/public/ADTelemetry.h
```diff
@@ -76,15 +57,15 @@
...
     Get a singleton instance of ADTelemetry.
  */
-+ (ADTelemetry*)sharedInstance;
++ (nonnull ADTelemetry*)sharedInstance;
 
 /*!
     Register a telemetry dispatcher for receiving telemetry events.
     @param dispatcher            An instance of ADDispatcher implementation.
     @param aggregationRequired   Specifies if telemetry events will be aggregated on client, 
                                  i.e. all events of a single request will be aggregated as one single event.
```
eddiekim: The comments for `aggregationRequired` are a little confusing. Why would I want aggregation? It sounds like a good thing until you see that the delegate interface is a bag of properties rather than being structured around an "event" or "events".

ADAL/src/public/ADTelemetry.h
```diff
@@ -76,15 +57,15 @@
 /*!
```
eddiekim: (from the comments above)

> Telemetry events will be flushed when function flush is called by developer.

Does the app developer need to call flush? Not clear how you do that from this interface.

eddiekim: whoops, wrong line...I meant here:

https://github.com/AzureAD/azure-activedirectory-library-for-objc/pull/859/files#diff-53fd7a8529454913eb5701503ea1a067R49

ADAL/src/public/ADTelemetry.h
jasoncoolmax: comment has been made clearer, I am gonna push the changes :)

ADAL/src/public/ADTelemetry.h
eddiekim: hmm ok...you'd still expect the delegate interface to be a little more structured around an "event" though. The dictionary interface is so open ended that the developer has no idea what they're going to get.

The parallel would be something like `ADUserInformation`. Yes, you could pass back all the information about a user with a dictionary (because that's all that we need), but it's nice to work with something typed.

ADAL/ADAL.xcodeproj/project.pbxproj
```diff
@@ -1101,6 +1103,7 @@
 				9453C3BE1C583AE6006B9E79 /* ADLogger.h */,
 				9453C3BF1C583AE6006B9E79 /* ADTokenCacheItem.h */,
 				6004019F1D340B760020EAAB /* ADTelemetry.h */,
+				605C167B1D8FBA590097DBA0 /* ADTelemetryEventStrings.h */,
```
RPangrle: I believe the edict is that this should be private.

RPangrle:
:shipit:

PR #866, opened 2017-01-05
Make ADTelemetry.h private again

RPangrle:
There's enough design talk on this feature still that we need to spend more time staging it, and can't hold up the larger 2.3.0 release for it.


jasoncoolmax:
:shipit:

PR #860, opened 2016-12-21
AD FS authority validation + testing + refactoring

unpluggedk:

PR #867, opened 2017-01-06
Remove 32-bit Mac and non-ARC support

### LINKED ISSUE #755: Remove 32-bit Mac Support
I don't think anyone is using it anymore now that Office is 64-bit only on Mac.

RPangrle:
There is one customer we added 32-bit support for has since gone 64-bit only. macOS has run only on 64-bit machines since 10.7 came out 5 years ago, so it's time to drop 32-bit Mac support.

Fixes #755


unpluggedk:
The following files are removed in #860 
ADAL/src/ADInstanceDiscovery.h, .m, 
ADAL/tests/ADInstanceDiscoveryTests.m 

ADAL/tests/ADAuthorityValidationTests.m replaces ADInstanceDiscoveryTests.m and still contains the changes applied in this PR. 

Please remove the first 2 mentioned and apply the removal of the non-ARC support on the third. Otherwise, please proceed and open an issue for me to follow up on,

Thanks!

RPangrle:
I removed the SAFE_ARC macros themselves, so I'll have to fix up any extra files before I can check in.

ADAL/src/cache/ADTokenCacheItem+Internal.h
```diff
@@ -27,12 +27,14 @@
 @class ADAuthenticationError;
 @class ADAuthenticationResult;
 
-@interface ADTokenCacheItem (Internal)
+@interface ADTokenCacheItem ()
+
+@property (readwrite) NSMutableDictionary * additionalClient;
+@property (readonly) NSDictionary * additionalServer;
```
jasoncoolmax: why do we put these two in class extension but the rest in class category? :)

ADAL/src/request/ADAcquireTokenSilentHandler.m
```diff
@@ -80,19 +79,10 @@ - (void)getToken:(ADAuthenticationCallback)completionBlock
 
 - (void)dealloc
 {
```
jasoncoolmax: shall we remove the dealloc fun as you did in other places? :)

ADAL/src/ui/mac/ADAuthenticationViewController.m
```diff
@@ -45,6 +45,13 @@ static NSRect _CenterRect(NSRect rect1, NSRect rect2)
 
 
 @interface ADAuthenticationViewController ( ) <WebResourceLoadDelegate, WebPolicyDelegate, WebFrameLoadDelegate, NSWindowDelegate>
+{
+    // In the legacy ObjC runtime (which is what we're stuck at for 32-bit Mac builds)
+    // you can't define the ivars of a class in the implementation file.
```
jasoncoolmax: do we still need these comments? :)

ADAL/src/ADInstanceDiscovery.m
```diff
@@ -0,0 +1,369 @@
+// Copyright (c) Microsoft Corporation.
```
jasoncoolmax: why do we still need this file? :)

ADAL/src/ui/mac/ADAuthenticationViewController.m
RPangrle: Possibly not.

ADAL/src/cache/ADTokenCacheItem+Internal.h
RPangrle: You can only add properties in an extension, unless you provide some other mechanism for their storage. Adding them like this means they'll get auto-synthesized and I don't have to bother providing implementation for them. :)

ADAL/src/ADInstanceDiscovery.m
RPangrle: No, it was supposed to be deleted in the merge :-/

ADAL/src/cache/ADTokenCacheItem+Internal.h
jasoncoolmax: Oh, got it ;)

ADAL/src/ADInstanceDiscovery.m
jasoncoolmax: Do we need to delete ADInstanceDiscoveryTest.m as well? :)

ADAL/tests/ADInstanceDiscoveryTests.m
```diff
@@ -0,0 +1,296 @@
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
```
jasoncoolmax: Do we need to delete ADInstanceDiscoveryTests.m as well? :)

RPangrle: 👍

PR #868, opened 2017-01-06
Push 1.2.x updates onto the ADAL cocoapod

RPangrle:
So we can (eventually) deprecate the ADALiOS cocoapod.


jasoncoolmax:
i signed off :)

Issue #483, opened 2016-02-01
Replace NSURLConnection with NSURLSession


RPangrle:
NSURLConnection was deprecated in iOS 9.

RPangrle:
Looking at the NSURLSession API this is going to have to wait for a 2.2 at least. It's a pretty big change with a lot of potential side effects.

RPangrle:
With iOS 10 coming out Apple is going to be less-and-less concerned about breaking something that was deprecated in iOS 9. We should probably look at doing this soon.

unpluggedk:
This has been completed as a PR referenced above.

PR #877, opened 2017-01-21
Convert NSURLConnection to NSURLSession

unpluggedk:


unpluggedk:
addresses #483

ADAL/src/urlprotocol/ADURLSessionDemux.m
```diff
@@ -0,0 +1,234 @@
...
+    self = [super init];
+    if (self)
+    {
+        self->_configuration = [configuration copy];
+        
+        self->_taskInfoByTaskID = [[NSMutableDictionary alloc] init];
+        
+        self->_sessionDelegateQueue = [[NSOperationQueue alloc] init];
+        [self->_sessionDelegateQueue setMaxConcurrentOperationCount:1];
+        [self->_sessionDelegateQueue setName:@"ADURLSessionDemux"];
```
RPangrle: If this string is never reused then putting it inline is acceptable.

ADAL/src/urlprotocol/ADURLProtocol.m
```diff
@@ -28,15 +28,16 @@
...
 
-static NSString* kADURLProtocolPropertyKey = @"ADURLProtocol";
+static NSMutableDictionary *s_handlers      = nil;
+static NSString *s_endURL                   = nil;
+static ADTelemetryUIEvent *s_telemetryEvent = nil;
 
+static NSString *s_kADURLProtocolPropertyKey  = @"ADURLProtocol";
 
-static NSUUID * _reqCorId(NSURLRequest* request)
+static NSUUID *_reqCorId(NSURLRequest* request)
```
RPangrle: And what exactly do you gain from using a class method instead of a static helper function?

ADAL/src/urlprotocol/ADURLProtocol.m
```diff
@@ -160,108 +173,67 @@ - (void)startLoading
...
-    if (!redirectResponse)
+    
+    if (!response)
     {
         // If there wasn't a redirect response that means that we're canonicalizing
         // the URL and don't need to cancel the connection or worry about an infinite
         // loop happening so we can just return the response now.
         
-        return mutableRequest;
+        completionHandler(mutableRequest);
```
RPangrle: Shouldn't there be a return statement after this?

ADAL/src/urlprotocol/ADURLProtocol.m
```diff
@@ -275,35 +247,87 @@ - (NSURLRequest *)connection:(NSURLConnection *)connection
...
+        [self.client URLProtocolDidFinishLoading:self];
+    }
+    else if ([error.domain isEqual:NSURLErrorDomain] && error.code == NSURLErrorCancelled)
+    {
+        // Do nothing.  This happens in two cases - 
+        // during a redirect, and request cancellation via -stopLoading.
+    }
+    else
+    {
+        [self.client URLProtocol:self didFailWithError:error];
```
RPangrle: This would result in duplicative logging.

ADAL/src/urlprotocol/ADURLSessionDemux.m
```diff
@@ -0,0 +1,234 @@
...
+    return self;
+}
+
+- (NSURLSessionDataTask *)dataTaskWithRequest:(NSURLRequest *)request delegate:(id<NSURLSessionDataDelegate>)delegate
+{
+    NSURLSessionDataTask *task;
+    ADURLSessionDemuxTaskInfo *taskInfo;
+    
+    task = [self.session dataTaskWithRequest:request];
+    taskInfo = [[ADURLSessionDemuxTaskInfo alloc] initWithTask:task
```
RPangrle: I think the @synchronized is being used to protect the taskInfoByTaskID mutable dictionary here, I'm not seeing the reason why the other lines would need to be protected.

ADAL/src/urlprotocol/ADURLSessionDemux.m
```diff
@@ -0,0 +1,234 @@
...
+                                                      delegate:delegate];
+    @synchronized (self)
+    {
+        self.taskInfoByTaskID[@(task.taskIdentifier)] = taskInfo;
```
RPangrle: Could you accomplish this with an associated property instead?

ADAL/src/urlprotocol/ADURLSessionDemux.m
RPangrle: I think the @synchronized is being used to protect the taskInfoByTaskID mutable dictionary here, I'm not seeing the reason why the other lines would need to be protected.

ADAL/src/urlprotocol/ADURLProtocol.m
RPangrle: There is already error logging when they are created or processed by the underlying code, adding extra logging here will result in duplicative logging.

ADAL/src/urlprotocol/ADURLSessionDemux.m
msolivialuo: It is used in a couple of lines below again.

ADAL/src/urlprotocol/ADURLSessionDemux.m
msolivialuo: When this method is executed on two different threads, one thread is updating task with different request (task = [self.session dataTaskWithRequest:request];) and the other is setting value in taskInfoByTaskID dictionary, you end up with the wrong task being set in the dictionary for the thread setting the value in dictionary.

ADAL/src/urlprotocol/ADURLSessionDemux.m
unpluggedk: It is being used at exactly 2 places, and they are separated by 2 lines.
I am open to making this constant but not convinced it adds any benefit here.

ADAL/src/urlprotocol/ADURLProtocol.m
unpluggedk: This has a void return type and this is the very last statement. I don't see a need to have a return after this.

RPangrle: If you don't return after this then all of the code after this if block will continue executing and your completion handler will be hit twice.

ADAL/src/urlprotocol/ADURLSessionDemux.m
RPangrle: Where is session being overridden? Where does it say that NSURLSession is not thread safe?

ADAL/src/urlprotocol/ADURLSessionDemux.m
```diff
@@ -0,0 +1,234 @@
...
+    
+    return task;
+}
+
+- (ADURLSessionDemuxTaskInfo *)taskInfoForTask:(NSURLSessionTask *)task
+{
+    ADURLSessionDemuxTaskInfo *result;
+    @synchronized (self)
+    {
+        result = self.taskInfoByTaskID[@(task.taskIdentifier)];
```
RPangrle: For future reference, there's no particularly compelling reason to not just return right away in the @synchronized block.

ADAL/src/urlprotocol/ADURLProtocol.m
unpluggedk: my bad, was looking at a wrong line.
addressed.

ADAL/src/request/ADWebRequest.m
```diff
@@ -162,95 +168,84 @@ - (void)send
...
+        //       dependent on the the challenge processing that the application performs.
+        //
+        NSAssert( _response != nil, @"No HTTP Response available" );
+        
+        ADWebResponse* response = [[ADWebResponse alloc] initWithResponse:_response data:_responseData];
+        [self completeWithError:nil andResponse:response];
+    }
+    else
+    {
+        [self completeWithError:error andResponse:nil];
```
unpluggedk: Let's add this to the things to do. 
#884

ADAL/src/urlprotocol/ADNTLMHandler.m
```diff
@@ -63,46 +63,57 @@ + (void)resetHandler
...
+             {
+                 NSURLCredential *credential;
+                 credential = [NSURLCredential
+                               credentialWithUser:username
+                               password:password
+                               persistence:NSURLCredentialPersistenceForSession];
+                 
+                 completionHandler(NSURLSessionAuthChallengeUseCredential, credential);
+                 
+                 AD_LOG_INFO_F(@"NTLM credentials added", nil, @"host: %@", challenge.protectionSpace.host);
```
unpluggedk: #884

ADAL/src/urlprotocol/ADNTLMHandler.m
```diff
@@ -63,46 +63,57 @@ + (void)resetHandler
...
+             } else {
```
unpluggedk: addressed.

ADAL/src/urlprotocol/ADURLProtocol.m
unpluggedk: no need this to be a class method if that is what you mean.
leaving as is.

ADAL/src/urlprotocol/ADURLProtocol.m
```diff
@@ -275,35 +247,87 @@ - (NSURLRequest *)connection:(NSURLConnection *)connection
...
 {
-    (void)connection;
+    NSString *authMethod = [challenge.protectionSpace.authenticationMethod lowercaseString];
+    AD_LOG_VERBOSE_F(@"session:task:didReceiveChallenge:completionHandler", _correlationId,
+                     @"%@. Previous challenge failure count: %ld", authMethod, (long)challenge.previousFailureCount);
     
-    [self.client URLProtocol:self didReceiveResponse:response cacheStoragePolicy:NSURLCacheStorageNotAllowed];
+    BOOL handled = NO;
+    Class<ADAuthMethodHandler> handler = nil;
+    @synchronized ([self class]) {
```
unpluggedk: addressed. thanks

ADAL/src/urlprotocol/ADURLProtocol.m
unpluggedk: thanks. leaving as is.

ADAL/src/urlprotocol/ADURLSessionDemux.m
```diff
@@ -84,21 +84,17 @@ @interface ADURLSessionDemux() <NSURLSessionDataDelegate>
 
 @implementation ADURLSessionDemux
 
+static const void *taskKey = &taskKey;
```
RPangrle: Rename this to `s_taskKey`

ADAL/src/request/ADWebRequest.m
```diff
@@ -167,10 +167,6 @@ - (void)send
     
     _task = [_session dataTaskWithRequest:request];
     [_task resume];
-
-//    _connection = [[NSURLConnection alloc] initWithRequest:request delegate:self startImmediately:NO];
-//    [_connection setDelegateQueue:_operationQueue];
-//    [_connection start];
```
RPangrle: 👍

PR #879, opened 2017-01-23
Remove 5-min timeout timer and increase webview connection timeout interval

### LINKED ISSUE #862: Remove the 5-min timeout timer before popping up webview
Currently there is a 5-min timeout timer set up before popping up webview for UI interaction. We may want to remove it completely as some logins may take quite some time.

jasoncoolmax:
fixes #862, #876

In order to fix the -1001 timeout issue, the following changes have been made:

1. The 5-min timeout timer has been removed before popping up webview.
2. The timeout interval of webview network connection has been increased.

The way to 
verify Change 1: stay at any login page for more than 5 mins.
verify Change 2: pick up the phone auth call, stay on the call for more than 1 min.

### DIFF
--- a/ADAL/src/ui/ADWebAuthController.m
+++ b/ADAL/src/ui/ADWebAuthController.m
@@ -568,7 +568,7 @@ static ADAuthenticationResult* s_result = nil;
     
     NSMutableURLRequest* request = [[NSMutableURLRequest alloc] initWithURL:[ADHelpers addClientVersionToURL:startURL]];
     [ADURLProtocol addCorrelationId:_requestParams.correlationId toRequest:request];
-    [request setTimeoutInterval:[[ADAuthenticationSettings sharedInstance] requestTimeOut]];
+    request.timeoutInterval = ADAuthenticationSettings.sharedInstance.requestTimeOut;
     
     [_authenticationViewController startRequest:request];
 }
--- a/ADAL/src/ui/ADWebAuthController.m
+++ b/ADAL/src/ui/ADWebAuthController.m
@@ -568,6 +568,8 @@ static ADAuthenticationResult* s_result = nil;
     
     NSMutableURLRequest* request = [[NSMutableURLRequest alloc] initWithURL:[ADHelpers addClientVersionToURL:startURL]];
     [ADURLProtocol addCorrelationId:_requestParams.correlationId toRequest:request];
+    [request setTimeoutInterval:[[ADAuthenticationSettings sharedInstance] requestTimeOut]];
+    
     [_authenticationViewController startRequest:request];
 }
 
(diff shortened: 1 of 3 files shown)


ADAL/src/urlprotocol/ADURLProtocol.m
```diff
@@ -139,7 +140,10 @@ + (NSURLRequest *)canonicalRequestForRequest:(NSURLRequest *)request
 {
     AD_LOG_VERBOSE_F(@"+[ADURLProtocol canonicalRequestForRequest:]", _reqCorId(request), @"host: %@", [request.URL host] );
     
-    return request;
+    NSMutableURLRequest* mutableRequest = [request mutableCopy];
```
RPangrle: This change makes me nervous. Like, very nervous. Before this method effectively did nothing but log out a bit. Now it's creating whole new request objects. The interaction between WebKit and NSURLProtocol is not particularly well understood and varies from OS version to OS version. Also the purpose of canonicalRequestForRequest is to do URL canonicalization, which is not happening here so that adds to the nervousness that it won't have the desired effect, or worse break with an OS update.

Is it at all possible to fix this bug without changing ADURLProtocol at all? Can we do this in -startLoading instead?

ADAL/src/ui/ADWebAuthController.m
```diff
@@ -224,20 +224,6 @@ - (void)stopSpinner
     [_authenticationViewController stopSpinner];
 }
 
-
-- (void)failWithTimeout
-{
-    _loadingTimer = nil;
```
jasoncoolmax: Seems we do want to remove it as it causes more pain than joy :)

ADAL/src/ui/ADWebAuthController.m
```diff
@@ -604,6 +568,8 @@ - (void)start:(NSURL *)startURL
     
     NSMutableURLRequest* request = [[NSMutableURLRequest alloc] initWithURL:[ADHelpers addClientVersionToURL:startURL]];
     [ADURLProtocol addCorrelationId:_requestParams.correlationId toRequest:request];
+    [request setTimeoutInterval:[[ADAuthenticationSettings sharedInstance] requestTimeOut]];
```
jasoncoolmax: sure :)

ADAL/src/urlprotocol/ADURLProtocol.m
jasoncoolmax: As we tested it, the 60s timeout can also come out from the request created by webKit. So canonicalRequestForRequest is the only place where we can modify that request :)

jasoncoolmax:
close as it needs to target 2.3.1

PR #881, opened 2017-01-25
Telemetry changes:

msolivialuo:
1. Added the telemetry rule class and defined rules for each event property
2. Move all the collection logics out from the event classes and added to the aggregator class.
3. Renamed all the property names to have _PROPERTY_NAME and property values to have _PROPERTY_VALUE in the name. This way the names are self-descriptive and easy to tell which one is property name vs. value.
4. Changed “#define” to actual function. “#define” should not be used as it is not type safe and difficult to debug.


ADAL/src/telemetry/ADTelemetryDefaultEvent.m
```diff
@@ -132,13 +125,6 @@ - (NSString*)getStringFromDate:(NSDate*)date
...
 }
 
-- (void)addAggregatedPropertiesToDictionary:(NSMutableDictionary*)eventToBeDispatched
-{
-    [eventToBeDispatched addEntriesFromDictionary:[self defaultParameters]];
-    SET_IF_NOT_NIL(eventToBeDispatched, AD_TELEMETRY_REQUEST_ID, [_propertyMap objectForKey:AD_TELEMETRY_REQUEST_ID]);
-    SET_IF_NOT_NIL(eventToBeDispatched, AD_TELEMETRY_CORRELATION_ID, [_propertyMap objectForKey:AD_TELEMETRY_CORRELATION_ID]);
-}
-
 - (NSDictionary*)defaultParameters
```
RPangrle: Make this a class method

ADAL/src/telemetry/ADTelemetryDefaultEvent.h
```diff
@@ -30,14 +30,15 @@
     NSInteger _defaultPropertyCount;
 }
 
+- (NSDictionary*)getProperties;
```
RPangrle: Assuming this is an accessor, remove 'get' from the method name.

msolivialuo: In the original code, this was a private method which returns a private dictionary. The change is to make the method public. I will keep it this way to avoid large changes.

ADAL/src/telemetry/ADTelemetryDefaultEvent.m
msolivialuo: This is a class method. Am I missing anything??

ADAL/src/telemetry/ADTelemetryDefaultEvent.h
msolivialuo: Updated.

ADAL/src/telemetry/ADTelemetryDefaultEvent.m
RPangrle: This got put on the wrong line somehow

ADAL/src/telemetry/ADTelemetryDefaultEvent.h
```diff
@@ -30,7 +30,7 @@
     NSInteger _defaultPropertyCount;
 }
 
-- (NSDictionary*)getProperties;
+@property (nonatomic, strong, readonly, getter=getProperties) NSDictionary *properties;
```
RPangrle: Getters should not have 'get' in them in objc

ADAL/src/telemetry/ADTelemetryDefaultEvent.m
RPangrle: This comment is on `-defaultParameters` which is an instance method. It should be a class method.

ADAL/src/telemetry/ADTelemetryEventStrings.h
```diff
@@ -34,76 +34,77 @@ extern NSString *const AD_TELEMETRY_EVENT_TOKEN_CACHE_WRITE;
...
-extern NSString *const AD_TELEMETRY_HTTP_ERROR_DOMAIN;
-extern NSString *const AD_TELEMETRY_AUTHORITY;
-extern NSString *const AD_TELEMETRY_AUTHORITY_AAD;
-extern NSString *const AD_TELEMETRY_AUTHORITY_ADFS;
-extern NSString *const AD_TELEMETRY_GRANT_TYPE;
-extern NSString *const AD_TELEMETRY_API_STATUS;
-extern NSString *const AD_TELEMETRY_EVENT_NAME;
-extern NSString *const AD_TELEMETRY_REQUEST_ID;
+
+extern NSString *const AD_TELEMETRY_PROPERTY_NAME_AUTHORITY_TYPE;
```
RPangrle: AD_TELEMETRY_PROPERTY_NAME => AD_TELEMETERY_KEY
AD_TELEMETRY_PROPERTY_VALUE => AD_TELEMETRY_VALUE

ADAL/src/utils/NSMutableDictionary+ADExtensions.m
```diff
@@ -0,0 +1,37 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+
+#import "NSMutableDictionary+ADExtensions.h"
+
+@implementation NSMutableDictionary (ADAL)
+
+- (void)setObjectWithNilCheck:(id)dictObject forKey:(id<NSCopying>)dictKey
```
RPangrle: Change to `-adSetObjectIfNotNil:forKey:`

ADAL/src/telemetry/ADTelemetryDefaultEvent.m
```diff
@@ -125,6 +125,19 @@ - (NSString*)getStringFromDate:(NSDate*)date
...
+- (void)sanitizePropertyNames
+{
+    NSMutableDictionary* propertyMapCopy = [_propertyMap copy];
+    [_propertyMap removeAllObjects];
+    
+    for (NSString* propertyName in propertyMapCopy)
+    {
+        // Some telemetry collectors only allow alphabetic characters, underscores, and dots in the property names
+        [_propertyMap setObject:[propertyMapCopy objectForKey:propertyName]
+                         forKey:[propertyName stringByReplacingOccurrencesOfString:@"-" withString:@"_"]];
```
RPangrle: If we're going to be not allowing hyphens in property names/keys then we should be changing the key values themselves.

msolivialuo: Checked with @jasoncoolmax before I made this change, these keys are not only used by telemetry, Changing these values will break existing apps.

msolivialuo:
Telemetry changes:
1. Added the telemetry rule class and defined rules for each event property
2. Move all the collection logics out from the event classes and added to the aggregator class.
3. Renamed all the property names to have _PROPERTY_NAME and property values to have _PROPERTY_VALUE in the name. This way the names are self-descriptive and easy to tell which one is property name vs. value.
4. Changed “#define” to actual function. “#define” should not be used as it is not type safe and difficult to debug.

PR #885, opened 2017-01-31
Copy authority rather than retain to avoid possible crash

jasoncoolmax:


ADAL/src/cache/ADTokenCacheAccessor.m
```diff
@@ -55,8 +55,7 @@ - (id)initWithDataSource:(id<ADTokenCacheDataSource>)dataSource
     _dataSource = dataSource;
     SAFE_ARC_RETAIN(dataSource);
     
-    _authority = authority;
-    SAFE_ARC_RETAIN(authority);
+    _authority = [authority copy];
```
jasoncoolmax: the known crash happened at line 75 if this file

RPangrle: This is too late. I meant to make sure in ADAuthenticationContext that we're not just retaining the passed in input, which we aren't. Doing a copy here is unnecessary.

jasoncoolmax: Make sense :)

jasoncoolmax:
not needed

PR #887, opened 2017-02-01
Not to set body if it is GET request

### LINKED ISSUE #886: Set web request to POST if body is set
Currently there is a bug in the code path that is only used by broker. Broker needs the fix for this.

jasoncoolmax:
fixes #886 

Broker needs this fix, otherwise there will lead to Get web request with body.
The bug  causes network requests failed to sent out in iOS 10.

For reference here:
ADAL 1.x:
https://github.com/AzureAD/azure-activedirectory-library-for-objc/blob/authenticator-hotfix/ADALiOS/ADALiOS/ADWebRequest.m#L67

ADAL 2.x
https://github.com/AzureAD/azure-activedirectory-library-for-objc/blob/dev/ADAL/src/request/ADWebRequest.m#L61"

---------------------------
Update:
We are taking another approach. We do not set body if a request is GET request

### DIFF
--- a/ADAL/src/request/ADWebAuthRequest.m
+++ b/ADAL/src/request/ADWebAuthRequest.m
@@ -75,8 +75,10 @@
         _requestURL = [NSURL URLWithString:newURL];
         SAFE_ARC_RETAIN(_requestURL);
     }
-
-    [self setBody:[[_requestDictionary adURLFormEncode] dataUsingEncoding:NSUTF8StringEncoding]];
+    else
+    {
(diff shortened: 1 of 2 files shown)


RPangrle:
Authenticator hotfix branch shouldn't exist anymore after the broker sdk refactor. Can this go into 2.3.1?

jasoncoolmax:
make sense :)

RPangrle:
@kpanwar Can you chime in here?

unpluggedk:
Linking an issue - #886

PR #896, opened 2017-02-14
Add custom headers in start loading

RPangrle:
NSURLConnection’s connection:willSendRequest:redirectResponse: was hit at the beginning of the request as well as redirects, while NSURLSession’s HTTP redirect is only hit on redirects. This was causing the initial request to not have a pkeyauth header, and thus, no pkeyauth challenge.


unpluggedk:
duplicate #897

PR #900, opened 2017-02-14
Fix documentation warnings

dtweston:
This PR fixes Xcode warnings like the following:

> `Parameter 'resource:' not found in the function declaration`


dtweston:
I also tried to turn on this warning for the project, so these issues can be caught by the maintainers in the future, but your setup with xcconfig looks a little complicated for me to understand.

RPangrle:
Thank you! I'll file an issue to fix up the xcconfigs as well.

dtweston:
Awesome, thanks for the prompt response!

PR #910, opened 2017-03-02
Change the logic to get parameters from the redirect url

msolivialuo:
When there are multiple colons in a url (for example: "urn:ietf:wg:oauth:2.0:oob?code=iamafakeauthenticationcode"), [NSURL query] will return nil. So we have to parse it and get the query parameters in the old way.


RPangrle:
The code you added back was removed because it was non-spec behavior. Considering this is a non-standard endpoint, we should avoid making non-spec changes for it.

ADAL/src/utils/NSString+ADHelperMethods.m
```diff
@@ -286,6 +286,27 @@ + (BOOL)adIsStringNilOrBlank:(NSString *)string
...
+                           start:(long)startIndex
+{
+    THROW_ON_NIL_ARGUMENT(set);
+    long end = self.length;
+    if (startIndex >= end)
+        return end;
+    
+    NSRange toSearch = {.location  = startIndex, .length = (end - startIndex)};
+    long found = [self rangeOfCharacterFromSet:set options:NSLiteralSearch range:toSearch].location;
+    return (found == NSNotFound) ? end : found;
```
partnerinflight: I'm curious about the contract here, and how come it'd be different from what rangeOfCharacterFromSet would return. i.e. why would you want to return self.length instead of NSNotFound if you can't find the character?

msolivialuo: I pulled the previous implementation to fix this. What I understood is that the caller is doing substringWithRange, if we use NSNotFound, which is max long, call to substringWithRange will go out of range.
Ideally this should be generic and let the caller to handle this case. Since it is well tested before, I will leave it this way for now. Does that make sense?

unpluggedk: I'd rather see this call return NSNotFound, and have the caller check for this condition.

msolivialuo: Since you both think this doesn't make sense, updated to return NSNotFound.

ADAL/src/utils/NSString+ADHelperMethods.m
```diff
@@ -292,11 +292,10 @@ - (long)adFindCharactersFromSet:(NSCharacterSet *)set
     THROW_ON_NIL_ARGUMENT(set);
     long end = self.length;
     if (startIndex >= end)
-        return end;
+        return NSNotFound;
```
unpluggedk: Missed braces

unpluggedk:
Test [NSURLExtensionsTests testQueryParameters] is failing

jasoncoolmax:
better to change the PR title, it is not related to NTLM but auth code end url parsing :)

PR #914, opened 2017-03-08
User NSURLComponents to get query instead of having our own parser.

msolivialuo:
1. Used NSURLComponents to get query instead of having our own parser.
2. Combined NSURL+ADExtension and NSURL+ADMethodsHelper to one class, no need to have multiple extension classes.
3. Added and updated UTs


ADAL/src/utils/ADNSURLComponentsHelper.h
```diff
@@ -21,11 +21,10 @@
...
 // THE SOFTWARE.
 
-
 #import <Foundation/Foundation.h>
 
-@interface NSURL (ADMethodHelpers)
+@interface ADNSURLComponentsHelper : NSObject
 
-- (BOOL)isEquivalentAuthority:(NSURL *)aURL;
++ (NSDictionary *)adQueryParameters:(NSURL *)url;
```
unpluggedk: No need to have 'ad' prefix, the reason was because it was an extension.

ADAL/src/utils/ADNSURLComponentsHelper.m
```diff
@@ -21,45 +21,17 @@
...
-    if (self.port || aURL.port)
-    {
-        if (![self.port isEqual:aURL.port])
-        {
-            return NO;
-        }
-    }
-    
-    return YES;
+    return [NSDictionary adURLFormDecode:components.query];
```
jasoncoolmax: just curious, any reason why you choose to put the implementation here rather than NSURL+ADExtensions? :)

msolivialuo: Good point indeed. Originally I thought NSURLComponents is different class. The extension/category classes can only be added to the same class. But we can call NSURLComponents internally in the method. Will update this.

PR #917, opened 2017-03-09
Automation app - add log element and move UI code to storyboard

unpluggedk:
- Add "adalLogs" to the result viewcontroller.
- move button to the top and remove keyboard logic (prev. through notifications)

- Simplify code by utilizing storyboard.
a. id of the elements visible through user defined runtime attributes
b. replacement of programatic constraints and embed within the storyboard


ADAL/ADALAutomation/ADAutoResultViewController.m
```diff
@@ -1,116 +1,53 @@
...
-    [self.view addConstraints:verticalConstraints];
-    [self.view addConstraints:horizontalConstraints];
-}
 
 - (void)viewDidLoad {
     [super viewDidLoad];
     // Do any additional setup after loading the view.
+    
+    self.resultInfo.text = self.resultInfoString;
+    self.resultLogs.text = self.resultLogsString;
```
mjcheetham: @unpluggedk, shouldn't the logs element be named `adalLogs`?
It should be consistent with the other automation test apps ([Android](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/weij/automationTestApp/automationtestapp/src/main/res/layout/activity_result.xml#L51), [Desktop](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/dev/automation/WinFormsAutomationApp/MainForm.Designer.cs#L45))

unpluggedk: There has been a recent talk about updating this to resultLogs for ADAL and MSAL. 
I will let you know after the discussion and when everything is settled.
Thanks!

ADAL/ADALAutomation/ADAutoRequestViewController.m
```diff
@@ -21,11 +21,46 @@
...
+#import "ADAutoRequestViewController.h"
+#import "ADLogger+Internal.h"
 
-@interface ADAutoTextAndButtonView : UIView
+@interface ADAutoRequestViewController ()
 
-@property (readonly) UIButton* actionButton;
-@property (readonly) UITextView* textView;
+@property (strong, nonatomic) IBOutlet UITextView *requestInfo;
+@property (strong, nonatomic) IBOutlet UIButton *requestGo;
```
partnerinflight: nit: consider "executeRequest" or something along those lines. "requestGo" sounds a bit weird

ADAL/ADALAutomation/ADAutoMainViewController.m
```diff
@@ -44,158 +64,185 @@ - (void)didReceiveMemoryWarning {
...
+        if(parameters[@"error"])
+        {
+            [self dismissViewControllerAnimated:NO completion:^{
+                [self displayResultJson:parameters[@"error"]];
+            }];
+            return;
+        }
+        
+        bool validateAuthority = YES;
+        if(parameters[@"validate_authority"])
```
unpluggedk: I mostly brought these over, and I generally agree with defines.

However, I think in this app is very much just for our automation testing, and thus, no need for retyping.
It is easier to just edit/view in place so it's more convenient for us.

ADAL/ADALAutomation/ADAutoMainViewController.m
```diff
@@ -44,158 +64,185 @@ - (void)didReceiveMemoryWarning {
...
+        {
+            validateAuthority = [parameters[@"validate_authority"] boolValue];
```
unpluggedk: Agreed. Thanks for the catch. It was mostly brought over from previous implementation 👍

ADAL/ADALAutomation/ADAutoRequestViewController.m
unpluggedk: The id of this button was asked to be requestGo, and I wanted the name to be in sync.

unpluggedk: Agreed, but the accessibilityIdentifier of this button was asked to be requestGo, and I wanted the name to be in sync.

ADAL/ADALAutomation/ADAutoMainViewController.m
unpluggedk: I mostly brought these over from previous implementation, and I generally agree with defines.

However, I think in this app is very much just for our automation testing, and thus,I believe the decision to do this was based on convenience of editing in place without jumping lines.

ADAL/ADALAutomation/ADAutoResultViewController.m
unpluggedk: @mjcheetham it's been confirmed that we will make everything "resultLogs"

mjcheetham: @unpluggedk, thanks for the confirmation. Do we have a time frame as to when the other automation apps will also be updated with the standardised names?

unpluggedk: I don't know of the time frame, all I know is 'eventually.'
I will see if I could get everyone's time on doing this change.

PR #925, opened 2017-03-23
Add login_hint for allowSilent request

jasoncoolmax:
Such that server won't return a different user just based on the cookies.

The if condition is copied over from line 109 from the same file.

(authenticator-hotfix-2.3.x branch is now the same as 2.3.1)

### DIFF
--- a/ADAL/src/request/ADAuthenticationRequest+WebRequest.m
+++ b/ADAL/src/request/ADAuthenticationRequest+WebRequest.m
@@ -245,6 +245,11 @@
             [requestData setObject:_scope forKey:OAUTH2_SCOPE];
         }
         
+        if ([_requestParams identifier] && [[_requestParams identifier] isDisplayable] && ![NSString adIsStringNilOrBlank:[_requestParams identifier].userId])
+        {
+            [requestData setObject:_requestParams.identifier.userId forKey:OAUTH2_LOGIN_HINT];
+        }
+        
         NSURL* reqURL = [NSURL URLWithString:[_context.authority stringByAppendingString:OAUTH2_AUTHORIZE_SUFFIX]];
         ADWebAuthRequest* req = [[ADWebAuthRequest alloc] initWithURL:reqURL
                                                               context:_requestParams];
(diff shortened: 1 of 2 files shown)


ADAL/src/request/ADAuthenticationRequest+WebRequest.m
```diff
@@ -245,6 +245,11 @@ - (void)requestCode:(ADAuthorizationCodeCallback)completionBlock
             [requestData setObject:_scope forKey:OAUTH2_SCOPE];
         }
         
+        if ([_requestParams identifier] && [[_requestParams identifier] isDisplayable] && ![NSString adIsStringNilOrBlank:[_requestParams identifier].userId])
```
jasoncoolmax: done

ADAL/tests/ADAcquireTokenTests.m
```diff
@@ -1650,4 +1666,45 @@ - (void)testResilencyTokenDeletion
...
+                                                     httpHeaderFields:@{}
+                                                     dictionaryAsJSON:@{}];
+    [ADTestURLConnection addResponse:response];
+    
+    // We send the actual silent network request
+    [req requestToken:^(ADAuthenticationResult *result)
+     {
+         // if request url is not expected,
+         // our network mock will fail the unit test and it won't hit here
+         XCTAssertNotNil(result);
```
partnerinflight: this is a nit, and feel free to ignore in the context of this review, but usually best practice unit testing discourages asserting on things you don't actually care about in the context of this specific unit test. In other words, if you have other UTs that assert that result is not null, you don't need to also assert that here. Assert only the things you are actually testing in this test. That helps keep the UTs less brittle -- if a fundamental assertion changes you don't have to fix 100 tests.

RPangrle: The thought here is because objc doesn't complain when you send a message to nil, it's possible for an assert to succeed even when the test should have failed. e.g.

```
XCTAssertEqual(result.status, 0)
```

Result could be `nil` and that assert would still pass because sending a message to nil also returns zero.  Checking for nils before checking result values can prevent tests passing that should have failed and make the reason why a test failed clearer.

RPangrle: The same principle holds up when you're writing objc code too. Even though it is safe to send messages to `nil` many-a-bug has been written based on this pattern:

```
if (![objectThatIsNil message]) { // do stuff that assumes it succeeded }
```

Issue #931, opened 2017-03-28
Telemetry typo: is_successfull


eddiekim:
`Microsoft.ADAL.is_successfull`

should be 

`Microsoft.ADAL.is_successful`

partnerinflight:
Since this is a public API, changing it now will involve cost to the host apps. Since it's a fairly benign spelling error, figure overall cost will be smaller leaving it as is.

PR #939, opened 2017-04-10
Fix telemetry typo

jasoncoolmax:
solves #931 

Question: do we want to make this fix to dev or 2.5.0 release branch?


jasoncoolmax:
Close it as it is not needed at this point of time.

PR #976, opened 2017-05-23
fix: Removed NSLog of the ADAL version in ADAuthenticationContext +load method

macblazer:
Every app launch logs the ADAL version; this is not helpful.  Every real log message coming out of the ADAL includes the ADAL version so there’s no reason to automatically log at app launch.


RPangrle:
This NSLog statement is very much on purpose. Not all applications implement the ADAL logger. Not all applications that implement the ADAL logger do so in a way that the logs are accessible. This allows us to quickly identify the version of ADAL used by simply running a binary without having to rely on other mechanisms.

PR #980, opened 2017-06-06
Define Input Files and Output Files on Info.plist shell phase

ksuther:
This makes it so that Info.plist doesn't have to get touched every
build, which can speed up projects that depend on ADAL (because the
framework doesn't have to be re-signed and re-copied every build).

The Info.plist file in the source tree is now modified instead of the
built product. This avoids problems with the New Build System option in
Xcode 9 where the shell script and the built-in file copy can cause
conflicts and result in the Info.plist being copied unnecessarily.


RPangrle:
Looks good, thanks for your contribution!

PR #1046, opened 2017-08-25
Move ADTestURLSession check into ADTestCase tearDown

RPangrle:


ADAL/tests/ADTestCase.m
```diff
@@ -40,6 +40,8 @@ - (void)setUp
 
 - (void)tearDown
 {
+    XCTAssertTrue([ADTestURLSession noResponsesLeft]);
```
antrix1989: @RPangrle I think tearDown is not a good place for a XCTAssert, because if it fails, Xcode does't highlight it inside a test and it is not obvious that assert failed during tearDown phase. I would say it is better to avoid XCTAssert in tearDown and put it  always inside a test.

PR #1070, opened 2017-09-21
Fix Xcode 9 issue that was causing codecov to fail in build script

RPangrle:
Xcode 9 changed the value returned for OBJROOT and the code coverage is no longer in that folder

Also added trace back printing on exceptions to make it a little easier to figure out what went wrong in the future.

### DIFF
--- a/build.py
+++ b/build.py
@@ -24,6 +24,7 @@
 # THE SOFTWARE.
 
 import subprocess
+import traceback
 import sys
 import re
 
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -221,14 +222,20 @@ def do_codecov(self) :
...
+			objroot = objroot[0:len(objroot) - 8]
+		codecov_dir = objroot + "/CodeCoverage"
 		executable_path = build_settings["EXECUTABLE_PATH"]
 		config = build_settings["CONFIGURATION"]
 		platform_name = build_settings.get("EFFECTIVE_PLATFORM_NAME")
 		if (platform_name == None) :
 			platform_name = ""
 		
 		command = "xcrun llvm-cov report -instr-profile Coverage.profdata -arch=\"x86_64\" -use-color Products/" + config + platform_name + "/" + executable_path
+		print command
```
unpluggedk: 👍

Issue #1100, opened 2017-11-16
Main Thread Checker: UI API called on a background thread: -[UIApplication canOpenURL:]


ggould75:
We just wanted to let you aware of this warning coming from the Main Thread Checker: 
**Main Thread Checker: UI API called on a background thread: -[UIApplication canOpenURL:]**
Not sure exactly from which version this started to appear, recently we updated to ADAL 2.5.2 (iOS)

```
...
WARNING: No cached preferred_network for authority
=================================================================
Main Thread Checker: UI API called on a background thread: -[UIApplication canOpenURL:]
PID: 95088, TID: 1277213, Thread name: (none), Queue name: com.apple.root.default-qos, QoS: 21
Backtrace:
4   ADAL                                0x000000011ba26505 +[ADBrokerHelper canUseBroker] + 182
5   ADAL                                0x000000011b9ff7d1 -[ADAuthenticationRequest(Broker) canUseBroker] + 85
6   ADAL                                0x000000011ba200b5 -[ADAuthenticationRequest(AcquireToken) requestTokenImpl:] + 55
7   ADAL                                0x000000011ba1fd81 -[ADAuthenticationRequest(AcquireToken) requestToken:] + 507
8   ADAL                                0x000000011ba1f990 -[ADAuthenticationRequest(AcquireToken) validatedAcquireToken:] + 439
9   ADAL                                0x000000011ba1f33e __70-[ADAuthenticationRequest(AcquireToken) acquireToken:completionBlock:]_block_invoke.141 + 412
10  ADAL                                0x000000011b9f4ed1 __74-[ADAuthorityValidation checkAuthority:validateAuthority:completionBlock:]_block_invoke + 138
11  libdispatch.dylib                   0x00000001105973f7 _dispatch_call_block_and_release + 12
12  libdispatch.dylib                   0x000000011059843c _dispatch_client_callout + 8
13  libdispatch.dylib                   0x00000001105a4499 _dispatch_root_queue_drain + 1444
14  libdispatch.dylib                   0x00000001105a3e97 _dispatch_worker_thread3 + 132
15  libsystem_pthread.dylib             0x0000000110a5b5a2 _pthread_wqthread + 1299
16  libsystem_pthread.dylib             0x0000000110a5b07d start_wqthread + 13
```

peng90:
[...]

PR #1130, opened 2018-01-12
Fix 'UI API called on a background thread' warning. #1100

antrix1989:


ADAL/src/broker/ios/ADBrokerHelper.m
```diff
@@ -185,6 +185,16 @@ + (void)load
 
 + (BOOL)canUseBroker
 {
+    if (![NSThread isMainThread])
+    {
+        __block BOOL result = NO;
+        dispatch_sync(dispatch_get_main_queue(), ^{
+            result = [self canUseBroker];
```
RPangrle: canOpenURL requires the main thread???? o.O

PR #1141, opened 2018-01-25
normalize userid for wipe call

unpluggedk:


unpluggedk:
This will create more conflicts when merging 2.6.x to dev

PR #1143, opened 2018-01-27
Update the user header search paths for the test apps

RPangrle:


RPangrle:
I had made the comment in that PR to fix it in xcconfigs, but I don't think @antrix1989 had gotten around to that yet.

PR #1148, opened 2018-02-01
Pass correlationId instead of context in ADURLProtocol.

antrix1989:


ADAL/ADAL.xcodeproj/project.pbxproj
```diff
@@ -3063,6 +3063,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = D6CF4EB11FC370BF00CD70C5 /* adal__framework__mac.xcconfig */;
 			buildSettings = {
+				GCC_OPTIMIZATION_LEVEL = 0;
```
RPangrle: Please don't check in build settings into the project file.

antrix1989:
Decided to put it on hold, since we will need to refactor ADRequestContext and it will affect the code across the project.

PR #1217, opened 2018-06-28
Add userId to error in non-broker flows

midella:


ADAL/src/request/ADAcquireTokenSilentHandler.m
```diff
@@ -152,7 +152,7 @@ - (void)acquireTokenByRefreshToken:(NSString*)refreshToken
          resultItem.authority = [_requestParams authority];
          
          
-         ADAuthenticationResult *result = [resultItem processTokenResponse:response fromRefresh:YES requestCorrelationId:_requestParams.correlationId];
+         ADAuthenticationResult *result = [resultItem processTokenResponse:response fromRefreshToken:(cacheItem ? cacheItem : nil) requestCorrelationId:_requestParams.correlationId];
```
jasoncoolmax: nit: should we simply pass `cacheItem` rather than `(cacheItem ? cacheItem : nil)`? :)

ADAL/src/request/ADAuthenticationRequest+WebRequest.m
```diff
@@ -66,7 +66,7 @@ - (void)executeRequest:(NSDictionary *)request_data
          item.clientId = [_requestParams clientId];
          item.authority = authority;
          ADAuthenticationResult* result = [item processTokenResponse:response
-                                                         fromRefresh:NO
+                                                    fromRefreshToken:nil
```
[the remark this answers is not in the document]
jasoncoolmax: I think it is ok to not return user id here. As we don't have a refresh token and we are not sure what user it is. Later after server returns the user information in the `response`, we could extract that out of the `response` :)

ADAL/src/cache/ADTokenCacheItem+Internal.m
```diff
@@ -88,9 +88,23 @@ - (ADAuthenticationResult *)processTokenResponse:(NSDictionary *)response
...
+            if (refreshToken.userInformation.userId)
+            {
+                [userInfo setObject:refreshToken.userInformation.userId forKey:ADUserIdKey];
+            }
+            error = [ADAuthenticationError errorWithDomain:error.domain
+                                                      code:error.code
+                                         protocolErrorCode:error.protocolCode
+                                              errorDetails:error.errorDetails
+                                             correlationId:requestCorrelationId
+                                                  userInfo:userInfo];
```
jasoncoolmax: Shall we have a function to handle the logic of adding entries to error userInfo? say `[ADAuthenticationError errorFromExistingError:additionalUserInfo:]`? :)

PR #1303, opened 2018-09-28
Hotfix/2.7.2 to dev

unpluggedk:


unpluggedk:
Will create master to dev instead

PR #1350, opened 2018-11-28
Update submodule.

antrix1989:


antrix1989:
ADAL returns on background thread during the interactive call (same behaviour as in version 2.6.5).

PR #1399, opened 2019-03-28
Flag to support disable/enable Kerberos.

unpluggedk:
common core PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-objc/pull/408


unpluggedk:
Re-doing this fix

PR #1494, opened 2019-12-09
Spelling correction in error message name.

dilipptdr:


antrix1989:
We cannot change the public API without a  major version bump.

PR #1577, opened 2021-08-23
Release/5.0.0

hieunguyenmsft:
Naming collision ADLogger are both used by ADAL and Apple's Depth (in ARKit) that makes app crash on iOS 15.
We change the name to ADALLogger, which requires Public API change.


kaisong1990:
Noticed the major version bumped, did we get agreement with stakeholders?

hieunguyenmsft:
> Noticed the major version bumped, did we get agreement with stakeholders?

I chatted with Olga about that, the public API is changed so according to the semantic version we need to bump major

antrix1989 (approved):

PR #1578, opened 2021-09-17
Remote PkeyAuthToken support for iOS simulator

antrix1989:


shakul78 (approved):
Better title - Remote PkeyAuthToken support for iOS simulator.

Thanks for committing the change as I was not able to do it due to permission issues.

PR #1592, opened 2022-04-06
Release/6.0.2

juan-arias:


mipetriu (commented):
Thanks, Juan. Do we need to update the version in ADAL.podspec manually?

juan-arias:
> Thanks, Juan. Do we need to update the version in ADAL.podspec manually?

I missed updating that file, it's updated now.

mipetriu (approved):

